PR #2, opened 2021-05-24
__memset fix

mrk-its:
After recent changes compilation of program using `printf` fails with:

```
ld.lld: error: undefined symbol: __memset
>>> referenced by ld-temp.o
>>>               lto.tmp:(_ntoa_long)
>>> referenced by ld-temp.o
>>>               lto.tmp:(_ntoa_long)
clang-13: error: ld.lld command failed with exit code 1 (use -v to see invocation)
```

It seems __memset symbol gets stripped from libc.a for some reason. I've found this solution here: https://www.keil.com/support/man/docs/armcc/armcc_chr1359124978363.htm - it solves it for me.

### DIFF
--- a/common/include/string.h
+++ b/common/include/string.h
@@ -15,7 +15,7 @@ void* memset(void *ptr, int value, size_t num);
 // Version of memset with better arguments for MOS. All non-pointer arguments
 // can fit in registers, and there is no superfluous return value. Compiler
 // intrinsic memset calls use this version, and user code is free to as well.
-void __memset(char *ptr, char value, size_t num);
+void __memset(char *ptr, char value, size_t num) __attribute__((used));
 
 #ifdef __cplusplus
 }


mrk-its:
I've just recompiled with fresh llvm-mos and it ofc works. Closing.

mysterymath:
Ah yes, this was due to me changing llvm-mos to use __memset instead of memset. This is technically a fix, since the calling conventions of the two differ, but LLD actually special-cases libcalls like memset to ensure that they're always included in link-time-optimization, even if they're not directly referenced. This is because generating code at link-time can spontaneously produce libcalls like memset, so they don't appear in the used symbols list until *after* the link has occurred. I fixed this by including "__memset" explicitly in the list of pseudos for our copy of LLD; I'll do so for any additional libcalls we introduce going forward.

PR #4, opened 2021-06-18
A800 chrout bugfix

mrk-its:
Current version of atari __chrout  doesn't work when inlined.


atari/800xl/lib/chrout.c
```diff
@@ -1,12 +1,14 @@
 void __chrout(char c) {
```
mysterymath, line 1: Instead, just "__attribute__((noinline)) void __chrout(char c) {"

This saves the extra overhead and is clearer about the intent.

mrk-its: done, works like a charm!

PR #5, opened 2021-06-19
Implement clock / reset_clock for sim

mrk-its:


sim/lib/stdlib.c
```diff
@@ -19,3 +19,17 @@ void exit(int status) {
 
   __builtin_unreachable();
 }
+
+unsigned long clock() {
+  // reading first byte latches whole 32-bit value
+  // so it needs to be read byte by byte in correct order
+  unsigned long ticks;
+  for(int i=0; i < 4; i++) {
+    ((unsigned char *)&ticks)[i] = *((volatile unsigned char *)0xFFF0 + i);
```
mysterymath, line 28: nit: just char, here and throughout. I made char unsigned for efficiency, so while unsigned char means the same thing, I don't want people to go spelunking into the SDK for "the best way to write llvm-mos C" and come away with the idea that they have to put unsigned char everywhere.

mysterymath, line 28: Just "char", here and throughout. I made char unsigned for efficiency, so while unsigned char means the same thing, I'm worried about people who go spelunking into the SDK for "the best way to write llvm-mos C" and come away with the idea that they have to put unsigned char everywhere.

mrk-its: probably size of `i` should also be reduced to `char`? Or llvm will do that automatically here?

mysterymath: That's one of the three "raison d'etre" optimizations of the compiler: the other being zero-page registers and statically-allocated stack frames. I'd leave it int, since it's more idiomatic.

mrk-its: oki, fixed.

mysterymath (approved):

PR #11, opened 2021-12-26
Add a simple heap implementation

karunski:
This is a small implementation of malloc/free/new/delete. 

It's implemented and standard new and delete, and it seems to integrate correctly with the default new and delete operators in c++ code.

Currently it just places a 4KB chunk of memory after the .noinit section.  The heap is "allocated" at the given address using  static variable initialization in the first alloc call, so it appears that if you only "pay" for the heap if you actually use it.  Otherwise it looks like the linker correctly optimizes it out.  For this to work I needed working "magic statics" in the c++ compiler, hence I implemented trivial `__cxa_guard_acquire` and `__cxa_guard_release` needed for llvm to generate the static initializer. 

As part of this I implemented some chunks of the freestanding c++ headers which are used within the heap implementation.


mysterymath:
Hi karunski!

Thanks for the PR! Presently, the libraries that come with LLVM-MOS are pretty minimal; we only have stub functionality necessary for the subset of LLVM's end-to-end test suite that we run. Even then, we don't have everything that these tests use, so quite a few of them are disabled.

This adds a ton of code to the SDK, and it would need a lot of work before inclusion. Having a version of malloc/new that generalizes well to the extremely heterogeneous environments that the 6502 finds itself in is a pretty hard problem, and one that we'd want to tackle before taking on the maintenance overhead a malloc.

karunski:
Can you elaborate on how I can proceed then?  I don't mind taking care of other things here in order to flesh out the SDK.

For example I don't see an end-to-end test suite that you refer to; hence I added the sample programs to demonstrate my heap works.

In a similar vein-- I'm not afraid of modifying / generalizing the heap implementation to work in more environments. I'm looking for feedback here on what you'd like in the malloc implementation. This PR is meant to be more of a starting point for that than an end state.

mysterymath:
> For example I don't see an end-to-end test suite that you refer to; hence I added the sample programs to demonstrate my heap works.

We maintain a fork of LLVM's end-to-end test suite here: https://github.com/llvm-mos/llvm-test-suite

At present, this runs about 2500 tests that give us some baseline confidence that the compiler as a whole is standards compliant. Presently, large swaths of it are foricbly disabled in the CMakeLists.txt. In particular, any tests that involved malloc or new had to be disabled.

Part of including malloc would be to uncomment those tests and get the onces that are appropriate on the 6502 passing. This would give us a degree of confidence that the malloc implementation is correct, complete, and standards compliant. We've also taken up the practice of modifying tests to make them reasonable on the 6502, wherever that's relatively straightforward.

Next, this includes a rather wide swath of headers for C++. I haven't gone through the C++ standard to any degree of detail, so it's difficult to verify whether or not these headers are precisely standards compliant. We'd probably generally be safer incorporating portions of an existing C++ standard library, since the C++ standard only fits on a rather large bookshelf from what I've heard. In any case, I don't have the means to verify whether these are standards compliant or not, so I can't really approve them without a lot of additional study.

There's also some issues on the C side of things; the new stdlib file clobbers existing abort functions in the SDK, for example. (This is the one that jumped out, but I didn't dig super far into this.)

Given the nature of target-specific memory maps on 6502 target machines, a malloc implementation should be capable of querying the underlying platform OS ROM to obtain the memory region available (where appropriate), or otherwise allow a target to somehow specify what memory region can be used for this. This would need to interact well with the memory used by the program itself for its sections. Alternatively, it'd be nice for a programmer to be able to limit the amount of memory available to malloc. I've a TODO to look at cc65 and see what they've done in this space; but there may be opportunities to do things a bit better.

Finally, I haven't had time to look at the malloc implementation itself, but it's concerning that there isn't any documentation regarding its operation, performance characteristics, or overall design. Dynamic memory allocator design has been continuously under development since the 70s, and given the tight performance bounds on 6502 code, we'd want to take advantage of this wherever possible. You may have written a modern allocator, but given the limited time I've allocated to this review, I haven't been able to dig through it and find out yet.

karunski:
So; my original motivation for this was to flesh out the c++ implementation, since the c++ code generation seems to work perfectly and most of the missing pieces were library related. For 6502 targets, it seems like the the freestanding c++ standard is the most reasonable c++ implementation to target. The set of headers needed to implement it is quite small; as you can see here: https://en.cppreference.com/w/cpp/freestanding.   Notably, the only portions of the library that have to be implemented are new/delete implementation, and some exception support. Exception support looked like it needed more help from the compiler, and was of dubious value to 6502 targets, so I decided to have a go a new/delete (and get malloc out of it as a side effect).

In my opinion, it's going to be easier to start from the ground up on the freestanding portion of c++ rather than trying to trim down an existing c++ stdlib (i.e. libc++ or similar).   In other words, there's more stuff to remove from a full implementation than there is to write in a from-scratch freestanding implementation. Feel free to provide counter-examples, though.

I missed that there was already an abort() implementation in there somewhere; it definitely makes sense to call that rather than reimplement it.

It's certainly fair to have the standard test suite of malloc family functions passing. I'll take a look at them and see what I can get working. 

As far as memory allocation and placement: traditionally the heap is just the section of address space between the end of statically declared variables (.bss etc) and the stack. The heap grows upwards to fill the space and the stack grows down, and you hope they don't run into each other.  As far as I can tell this is what is done by cc65 and other compilers that target a machine without an OS with virtual memory. (i.e. freestanding implementations)

For the C64 it looks like this is all that's needed because the space between our current stack impl and the end of .bss is all the free memory.  I could dynamically set the heap size here if I knew how big the stack could grow-- but I didn't think that was available at program start?   Is there any way to limit the stack size?

For Atari 8-bit it looks like the same general idea-- it looks like the current linker script just chooses a safe memory start point, the program is placed there, the heap start is placed after that, the stack is placed at the end of memory.  

For NES, there's only 2K doesn't seem like it would have enough room to bother with trying to squeeze in a heap implementation at all. 

Beyond that; I know there exist systems that allow access beyond the 64k address space of the 6502; but as far as I can tell these all require the program to explicitly control paging / bank switching. I don't think those details can be abstracted away by the heap implementation; they require explicit programmer control over paged regions of memory.  

Currently I've limited the heap to just 4K of RAM.  I could change that to a compile-time defined constant, or perhaps pick a heap size that actually intelligently fills everything below the "top of the stack".  Either way; as far as I can see this is a static maximum size that is set at program start. The ROM or OS, such as it is, isn't going to change the amount of memory available to the program.  All the memory that exists in the system is available at load time.

The heap implementation is a first-fit free list (https://en.wikipedia.org/wiki/Free_list) inspired in part by this implementation https://embeddedartistry.com/blog/2017/02/15/implementing-malloc-first-fit-free-list/. Certainly more heap allocators are possible, but with only a few kbytes to work with the code size seems like it would produce diminishing returns: for example a pooling allocator might save on average a few bytes per allocation, but the implementation of the heap itself might take a few more kB of code that you'll never make up for with allocation savings. 

The nice thing about a free list implementation is that the poor locality of reference doesn't really matter for this target since AFAIK 6502 never had any cache. It seems like the best default choice for 6502.

mysterymath:
> So; my original motivation for this was to flesh out the c++ implementation, since the c++ code generation seems to work perfectly and most of the missing pieces were library related. For 6502 targets, it seems like the the freestanding c++ standard is the most reasonable c++ implementation to target. The set of headers needed to implement it is quite small; as you can see here: https://en.cppreference.com/w/cpp/freestanding. Notably, the only portions of the library that have to be implemented are new/delete implementation, and some exception support. Exception support looked like it needed more help from the compiler, and was of dubious value to 6502 targets, so I decided to have a go a new/delete (and get malloc out of it as a side effect).

Ah, this makes sense as to why this was targeted first. I don't think this is actually all that's necessary though; IIRC clang++ expects the Itanium C++ ABI (or something like that) to be present; you ran into one of these with `__cxa_guard` or whatever it was. I think there's actually quite a few of these, particularly to do with the vtable manipulation and the runtime support needed to run code before main (constructors of statically initialized C++ objects).

> In my opinion, it's going to be easier to start from the ground up on the freestanding portion of c++ rather than trying to trim down an existing c++ stdlib (i.e. libc++ or similar). In other words, there's more stuff to remove from a full implementation than there is to write in a from-scratch freestanding implementation. Feel free to provide counter-examples, though.

I think this is generally true, with the possible exception of stuff like type traits and other generic headers. The header-only stuff should be more or less identical, and it saves trying to go through the POSIX or C++ standards clause by to make sure it's up to snuff (as I've done for everything presently in the SDK).

> For the C64 it looks like this is all that's needed because the space between our current stack impl and the end of .bss is all the free memory. I could dynamically set the heap size here if I knew how big the stack could grow-- but I didn't think that was available at program start? Is there any way to limit the stack size?
> 
> For Atari 8-bit it looks like the same general idea-- it looks like the current linker script just chooses a safe memory start point, the program is placed there, the heap start is placed after that, the stack is placed at the end of memory.
> 
> For NES, there's only 2K doesn't seem like it would have enough room to bother with trying to squeeze in a heap implementation at all.

Ah, I think the disconnect may have actually been with the way we set up the stack for these targets, not the heap. My recollection of the programmer's reference manuals are hazy, but IIRC the OS of the Atari maintains a MEMTOP pointer that identifies the highest user-accessible memory region, based on what graphics modes are used. On that platform at least, we'd probably want the stack to start there and grow down, perhaps with a mechanism to allow code to run pre-main to initialize this. Or something.

I'm actually not sure what happens when graphics modes are changed on the Atari; that target was contributed by someone else. Very little design work has occurred on polishing these so far; there's just enough to verify that the compiler is prima facie portable, and little else.

> Currently I've limited the heap to just 4K of RAM. I could change that to a compile-time defined constant, or perhaps pick a heap size that actually intelligently fills everything below the "top of the stack". Either way; as far as I can see this is a static maximum size that is set at program start. The ROM or OS, such as it is, isn't going to change the amount of memory available to the program. All the memory that exists in the system is available at load time.

If we could manage it, I think it'd be pretty sensible to have a C function that changes the limit of the heap size. We wouldn't be able to really detect if the stack rams into the heap, but we could allow users to control how many mallocs can occur before it starts failing.

> The heap implementation is a first-fit free list (https://en.wikipedia.org/wiki/Free_list) inspired in part by this implementation https://embeddedartistry.com/blog/2017/02/15/implementing-malloc-first-fit-free-list/. Certainly more heap allocators are possible, but with only a few kbytes to work with the code size seems like it would produce diminishing returns: for example a pooling allocator might save on average a few bytes per allocation, but the implementation of the heap itself might take a few more kB of code that you'll never make up for with allocation savings.
> 
> The nice thing about a free list implementation is that the poor locality of reference doesn't really matter for this target since AFAIK 6502 never had any cache. It seems like the best default choice for 6502.

From Doug Lea's experience and the Wilson et al survey, the devil is in the details with allocator implementations. Slight tweaks to the data structures composing the free list can save a considerable amt of RAM, and IIRC his allocator only took around 13 machine instructions per malloc on average.

That being said, a simple malloc would be fine to start with, so long as we can verify it is correct and hack on it later to make improvements. We'll almost certainly want to rewrite the core path in assembly eventually, and binning is a very likely addition as well. We'll want to set up benchmarks that hammer at malloc before doing so; there are probably some relatively standardized approaches we can grab.

Accordingly, we'd need the code to be quite a bit better documented and/or variable/structure/object names more descriptive. It's a bit hard to follow as is (at least for my frazzled holiday-brain). If this was code that we could tuck away and never have to think about (so long as it works), I wouldn't actually care that much, but since this is malloc, we're invariably going to have to spend a lot of time measuring, tuning, tweaking, and generally just picking at it to get it fast and small.

Finally, all that said, I do think having real freestanding C++ would be a tremendous feather to add to the project's cap, so if you're interested, I'm willing to spend the extra time on PRs along those lines. I think you're right that there's not all that much to do to get there, and it'd be really killer to have both working around the time we do our first official release (whenever that is).

mysterymath:
Oh, and I should add that there are a bunch of C++ end-to-end tests in the test suite that we had to turn off because of missing C++ ABI calls; these should help distinguish which actually need to be implemented to get to a full freestanding C++ implementation.

karunski:
Updated as follows:

1. Turned on tests that I could find in llvm-test-suite.  See corresponding PR on that repo. 
2. Added control over size of the heap.
3. Updated docs with a description of the heap implementation and documentation on the non-standard heap size limiting functions.  (see stdlib.h)
4. Updated examples to demonstrate functionality; added an example toolchain file for building on c64 with the SDK.

mysterymath:
Just a heads up:
A pretty big rework of the SDK is currently underway: https://github.com/llvm-mos/llvm-mos-sdk/pull/12

Nothing is finalized yet, but it's pretty likely that the syntax used for defining SDK platforms in CMake will significantly change. I'll try to review the pure C++/C parts of this when I get the chance, but the remainder will probably need to be updated once that change is finalize and lands.

karunski:
OK; no problem.  I had a quick glance at PR #12 and it seems good.  I ran into some of the same problems that PR addresses; specifically CMake needs to be invoked with cross-compile toolchain functionality to work cleanly when targeting MOS. I kept fighting it's tendency to add flags that are normally required for my host platform; e.g. adding OSX_SYSROOT when using MacOS as my host environment.  

I'll rework / rebase this PR as needed to deconflict with those changes.

commodore/64/include/stdlib.h
```diff
@@ -0,0 +1,16 @@
...
+#define _C64_STDLIB_H_
+
+#include_next <stdlib.h>
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+    void abort();
```
mysterymath, line 11: nit: Align this all the way to the left.

commodore/64/lib/stdlib.c
```diff
@@ -0,0 +1,9 @@
+// place holder for stdlib "abort" function.
+
+// Not sure what the best way to handle this is on C64.
+
+void abort()
+{
+    asm volatile("brk" : );
```
mysterymath, line 7: Looking at the example set by cc65, this should probably actually return to BASIC, at least if the memory map is set up to keep the BASIC environment usable.

The behavior of BRK without a custom monitor or handler is pretty questionable though; an infinite loop is probably safer until we really get into the weeds of how this should work.

commodore/64/lib/stdlib.c
```diff
@@ -0,0 +1,9 @@
+// place holder for stdlib "abort" function.
+
+// Not sure what the best way to handle this is on C64.
+
+void abort()
+{
```
mysterymath, line 6: Braces on same line for function definitions, here and throughout the PR.

common/include/cstddef
```diff
@@ -0,0 +1,89 @@
...
+#include <type_traits>
+
+namespace std
+{
+using size_t = ::size_t;
+using ptrdiff_t = ::ptrdiff_t;
+using max_align_t = ::max_align_t;
+using nullptr_t = decltype(nullptr);
+
+enum class byte : unsigned char
```
mysterymath, line 13: Braces on same line as enum/struct/class definitions; here and throughout the PR.

common/include/cstddef
```diff
@@ -0,0 +1,89 @@
...
+}
+
+template <class IntegerType>
+constexpr enable_if_t<is_integral<IntegerType>::value, byte &>
+operator>>=(byte& b, IntegerType shift) noexcept
+{
+    return b = b >> shift;
+}
+
+}
```
mysterymath, line 87:
```
} // namespace std

#endif // _CSTDDEF_ 
```

commodore/64/include/stdlib.h
```diff
@@ -0,0 +1,16 @@
+#ifndef _C64_STDLIB_H_
+#define _C64_STDLIB_H_
+
+#include_next <stdlib.h>
+
+#ifdef __cplusplus
+extern "C"
```
mysterymath, line 7: Opening brace on same line as `extern "C"`, here and throughout the PR.

common/include/cstdint
```diff
@@ -0,0 +1,38 @@
+#ifndef __CSTDINT_H__
+#define __CSTDINT_H__
+
+#include <stdint.h>
+
+namespace std
```
mysterymath, line 6: Opening brace on same line as namespace definition; here and throughout the PR.

mysterymath (changes requested):
Going through these file by file; these are comments up to, but not including, `common/include/new`.

common/lib/CMakeLists.txt
```diff
@@ -12,6 +12,8 @@ add_library(common_crt OBJECT
    divmod.cc
    mul.cc
    shift.cc
+
+   new.cc
```
mysterymath, line 16: As is, this doesn't really fit in the C runtime library, since it's not a part of the C runtime; it's a part of the C++ runtime.

For LLVM, implementations of most of these functions appear to be in their libcxx; these should probably be in a libc++ that forms the C++ runtime and standard library.

Generally, it's expected to have a libc++ depend on a libc, but not the other way around. Accordingly, we can move the actual heap implementation into libc, and only reference it from libc++. We can make the compiler driver  automatically import our libc++ if called via clang++, as is the usual practice. This keeps the C linker namespace clean of mangled C++ symbols like _ZNewWhatever.

This doesn't matter all that much, but it's generally important to carefully allocate a project's "surprise points." If there's a really specific reason to implement the heap in new.cc, let me know; we can then document that and have an explanation if folks come asking about it.

mysterymath (changes requested):
Another day, another round of comments :)

common/include/stdlib.h
```diff
@@ -9,6 +11,101 @@ int abs(int i);
...
+3. After the first allocation has been made, the heap may only increase in size.
+   Any attempt to decrease the size of the heap limit will be ignored.
+
+Increasing the heap limit does not do any validation to ensure the heap will not
+collide with the stack. You must leave enough space for whatever stack usage
+your program needs.
+*/
+
+/* Return the current maximium size of the heap.*/
+size_t heap_limit();
```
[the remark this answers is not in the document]
mysterymath: Took a quick look at what glibc did for its extensions, and it looks like there's an option 3:

3) Define these functions using the `__heap_limit()` reserved naming convention, then do:
```
#ifdef _MOS_SOURCE
#define heap_limit __heap_limit
#endif
```
Then, you can
```
#define _MOS_SOURCE
#include <stdlib.h>
```
To get access to the functions under their simple names.

common/include/new
```diff
@@ -0,0 +1,31 @@
...
+void *operator new(std::size_t size);
+void *operator new[](std::size_t size);
+void *operator new(std::size_t count, const std::nothrow_t &tag) noexcept;
+void *operator new[](std::size_t count, const std::nothrow_t &tag) noexcept;
+void *operator new(std::size_t count, void *ptr);
+void *operator new[](std::size_t count, void *ptr);
+void operator delete(void *ptr) noexcept;
+void operator delete[](void *ptr) noexcept;
+
+// void dump();
```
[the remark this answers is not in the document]
karunski: I just deleted the dump() code; it wasn't general enough to be useful to a library consumer.

common/lib/CMakeLists.txt
mysterymath: Makes sense; we do probably have some pre-work to do to establish a libc++. At the very least, I'll want to survey other compiler targets to get a feel for naming conventions; I've seen libcpp, libcppabi, libc++, libstdc++.

For now, if you would, move this stuff to libc. It definitely doesn't belong in libcrt (which is exclusively for compiler-internal libcalls), and we'll still probably want the malloc stuff to stay in libc when we get around to building a real libc++.

common/lib/new.cc
```diff
@@ -0,0 +1,369 @@
+#include <cstdint>
+#include <new>
+#include <stdio.h>
+
+extern std::byte __heap_start;
+
+namespace {
+
+// An arbitrary block of memory. This struct only holds the size of the block,
+// and the data in the block directly follows
```
mysterymath, line 10: nit: End sentences in periods, here and throughout.

examples/hello_malloc.c
```diff
@@ -0,0 +1,73 @@
+#include <stdint.h>
```
mysterymath, line 1: These two files currently play the role of both tests of malloc/new and examples for their usage.

As tests, they could be improved by incorporating them into our end-to-end test suite, replacing the printing with appropriate failure signaling (typically, `abort()`ing if anything goes wrong). This would allow us to get automatically notified if we break malloc.

As examples, they could be improved by simplifying the remainder down into something very simple: mallocing some memory, doing something with it, then freeing it. This would make it easier to quickly evaluate the runtime and code overhead of malloc by examining the generated output, which would help users understand whether or not it's a good fit for their designs.

common/lib/new.cc
```diff
@@ -0,0 +1,369 @@
...
+}
+
+void operator delete[](void *ptr) noexcept { operator delete(ptr); }
+
+// helper functions for getting/setting flags in guard_object
+static bool initializerHasRun(std::uint8_t *guard_object) {
+  return (*guard_object != 0);
+}
+
+static void setInitializerHasRun(std::uint8_t *guard_object) {
```
mysterymath, line 313: As I was scrolling around the llvm source, I found that this was copied from the first Apple Google search result, comments and all.

The Apache license requires that this sort of thing include the original copyright notice in the source file. Please either do this, or rewrite the copied portions in your own terms.

There's probably a fair use argument here, but such arguments have become increasingly fraught as of late, and it's easier and safer to just follow the rules.

common/lib/new.cc
```diff
@@ -0,0 +1,369 @@
...
+  *(guard_object) = 1;
+}
+
+extern "C" {
+int __cxa_guard_acquire(std::uint8_t *guard_object) {
```
mysterymath, line 318: This doesn't actually provide the guarantees that __cxa_guard_acquire is supposed to in the presence of asynchronicity.

Example:
1. The main program enters a static initializer, calling __cxa_guard_acquire. initializerHasRun completes, and returns zero.
2. Before this can return, an NMI interrupt fires. The NMI handler calls the same function. Since nothing has yet been written back, initializerHasRun also returns 0. This causes the whole initialization to run on the NMI thread.
3. The NMI returns. The original __cxa_guard_acquire returns 1, and the whole initialization runs a second time, this time on the main thread.

Disabling interrupts doesn't really fix this, since NMI can call C code. It's possible to disable NMIs on cerrtain platforms (maybe all?), but the means for doing so is platform specific.

There's also the possiblity that this would be runnint in a more hosted environment, say, one that provides real threads and async signals. In that case, a real lock implementation might be on offer, and that would be usable here.

Accordingly, we could make a platform-specific set of lock or atomic primitives and use those to implement these. There's a couple ways I could see doing that:
1. Disable all interrupts using platform-specific mechanisms. If we came up against a plaform where there was no way to disable NMI, we'd be SOL if this specific scenario occurs. Even though it's rare, it would create pointy bits in the user experience that we've mostly been able to avoid so far.
2. Find a clever way to use the uninterruptable INC and DEC operations to implement a kind of semaphore or lock. If there's a way to do this, we could have a truly generic lock/atomic implementation. I'd prefer this if possible, since it keeps the complexity away from the user facing side of things, and it generalizes across platforms.
3. Implement one of the fancy algorithms (https://cs.stackexchange.com/questions/30247/are-there-hardware-lock-implementations-without-test-and-set-or-swap) for locking over basic primitives. I haven't looked into any of these in detail, but if they work as they claim on the tin, then it should be a workable fallback. Not sure if any of them are *fast* though.

mysterymath (changes requested):

common/lib/new.cc
mysterymath: It occured to me after writing this that the situation is trickier for interrupt handlers than for traditional threads, since the interrupted control flow cannot resume until the interrupt returns. This means that any traditional lock held by an interrupted thread would deadlock if the same lock was required by the interrupting thread. For the static allocation case, if the main thread began initializing a function static variable, and an NMI needed access to the same static, there'd be absolutely no safe way to proceed. The NMI needs access to the static before it can proceed and return, but there's no way for the main thread to finish initializing it. The NMI can't safely initialize it either, since it's already in the process of being initialized.

This is something that will need more careful thought; perhaps an inspection of the behaviors of pre-existing freestanding C++ environments might be illuminating.

mysterymath: From https://sudonull.com/post/139020-C-support-for-avr-in-gcc, one way to deal with this is to completely disable interrupts for the duration of the initialization. That is, disable them in `__cxx_guard_acquire`, and reenable them in `__cxx_guard_release`. This would imply having a platform-specific interrupt disabling mechanism, though.

mysterymath: Oh, a cheesy way to disable NMI on any platform is just to point the system interrupt vector to an RTS for the duration, then restore it back. It's still not 100% ideal; if a static initializer or other process were to manipulate the NMI vector, then it would reenable interrupts, causing bad things. So we'd have to call out manipulating the NMI vector while a C++ static initialization is active as undefined behavior, which we're in our rights to do, since NMI manipulation can only be done via implementation-defined behavior.

mysterymath: Ah, but changing the NMI vector is also usually platform specific. Que sera.

mysterymath: Alllllternatively we could declare doing this specific thing in an NMI to be undefined behavior, just set the interrupt disable flag, and call it a day.

examples/hello_malloc.c
karunski: These have been simplified, and the "real tests" have been added to the test_suite PR.  To make this work I defined an __LLVM_MOS_SDK preprocessor #define so the tests could properly exercise the failed allocation paths and the heap limit functionality.

common/lib/new.cc
karunski: Right; I'll have to think on this. Will get back to you on this later.

karunski: OK-- I think you're over-thinking this.  This problem is not unique to freestanding implementation.  Hosted implementations also have interrupt service functions, for example consider signal handlers.  

The one-time construction API only provides *thread* safety, not *interrupt* safety.  If there is only a single thread of execution, we can reason that an interrupt handler cannot execute concurrently with the code that it interrupted. Therefore, any attempt for __cxa_guard_acquire to acquire the lock within the interrupt routine would fail/deadlock if the interrupt is triggered while the thread had already acquired the lock.

I am wagering that any implementation of __cxa_guard_acquire fails in the context of an interrupt service routine. I'm curious what happens if you try this on a regular, hosted clang++ implementation.  I'll try it.

I'm pretty sure disabling interrupts within the __cxa_guard_acquire is a non-starter, for any hosted or freestanding implementation. The one time initialization guard has to be invoked every time execution references the static variable so it has to be extremely cheap to test if the initialization has already occurred. Enabling/disabling interrupts every time you reference a static function scope variable would be both too expensive and too surprising. 

So; for this single-threaded implementation, the trivial non-implementation of __cxa_guard is probably the correct one.

mysterymath: So, the standard was a bit tricky to interpret here; the language used doesn't say anything about thread safety; it just says:

> If control enters the declaration concurrently while the variable is being initialized, the concurrent execution shall wait for completion of the initialization.


Which would, prima facie, include any means by which concurrent execution could occur.

However, independently of this, a function is not signal-safe if, amongst other things, it is "waiting for the completion of the initialization of a variable with static storage". Performing a non-signal-safe evaluation in a signal handler (which we can generalize to interrupt handlers) is undefined behavior. So the wording of the above becomes moot.

Accordingly, we can go ahead and resolve this. Please do add a comment above these functions summarizing the discussion; this should help casual readers understand why we don't need a more sophisticated __cxa_guard.

common/include/stdlib.h
```diff
@@ -9,6 +15,111 @@ int abs(int i);
...
+      │                            │    └───────────────────────────────────────┘  
+      │   Reserved lower memory:   │                                               
+      │         zero-page,         │                                               
+      │    hardware stack, etc.    │                                               
+      │                            │                                               
+      │                            │                                               
+0x0000└────────────────────────────┘                                               
+*/
+
+void *malloc(size_t size);
```
mysterymath, line 77: Please mark these allocator functions as weak, so OS/application developers can substitute their own custom malloc and free. This would in turn allow the standard library and new to automatically pick up the new allocator.

Along these lines, we really should try to move the implementation of the allocator into malloc itself; it'd be very surprising if a user malloc wasn't picked up by operator new() and operator new[].

common/lib/new.cc
```diff
@@ -0,0 +1,377 @@
...
+void *operator new(std::size_t size) {
+  const auto block = operator new(size, std::nothrow);
+  if (!block) {
+    run_new_handler();
+  }
+
+  return block;
+}
+
+void *operator new[](std::size_t size) { return operator new(size); }
```
mysterymath, line 289: Mark these as weakly linked (attribute(weak)); it's customary to do so to allow OS/app devs to provide custom implementations of the baseline new and delete.

common/lib/new.cc
```diff
@@ -0,0 +1,377 @@
...
+static bool initializerHasRun(std::uint8_t *guard_object) {
+  return (*guard_object != 0);
+}
+
+static void setInitializerHasRun(std::uint8_t *guard_object) {
+  *(guard_object) = 1;
+}
+
+extern "C" {
+int __cxa_guard_acquire(std::uint8_t *guard_object) {
```
mysterymath, line 326: Mark these functions as weak; that way an OS/application developer can provide their own versions that provide the appropriate guarantees in the presence of custom threading primitives.

mysterymath (changes requested):

common/include/cstdint
```diff
@@ -0,0 +1,37 @@
...
+using uint_fast64_t = ::uint_fast64_t;
+using uint_least8_t = ::uint_least8_t;
+using uint_least16_t = ::uint_least16_t;
+using uint_least32_t = ::uint_least32_t;
+using uint_least64_t = ::uint_least64_t;
+using uintmax_t = ::uintmax_t;
+using uintptr_t = ::uintptr_t;
+} // namespace std
+
+#endif
```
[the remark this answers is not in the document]
karunski: I think that the it's debatable whether an interrupt constitutes concurrent execution. In this scenario, it looks like other implementations consider this a recursion problem since the two entries into the __cxa_guard_acquire can't be concurrent.

It was interesting to see what clang and gcc do with the code below. Clang/libc++ aborts with `libc++abi.dylib: __cxa_guard_acquire detected recursive initialization`  and gcc (on linux) throws an undocumented internal exception type with a similar message.


```
#include <string_view>
#include <signal.h>
#include <unistd.h>
#include <stdio.h>

using namespace std::literals::string_view_literals;

int count_construction = 0;
volatile bool recvd_sigusr1 = false;

class Singleton
{
public:
	static Singleton & getInstance()
	{
		static Singleton instance;
		return instance;
	}

	bool is_initialized() const noexcept { return m_initialized; }
private:
	Singleton() noexcept
	{
		count_construction += 1;

		while (!recvd_sigusr1) {}

		m_initialized = true;
	}

	bool m_initialized = false;
};

void puts_signal_safe(std::string_view msg)
{
	::write(1, msg.data(), msg.length());
	::write(1, "\n", 1);
}

void on_usr1(int)
{
	recvd_sigusr1 = true;
	auto & instance = Singleton::getInstance();
	if (count_construction > 1)
	{
		puts_signal_safe("constructed more than once!"sv);
	}
	if (!instance.is_initialized())
	{
		puts_signal_safe("not initialized!"sv);
	}
}

int main(int argc, char * argv[])
{
	signal(SIGUSR1, on_usr1);

	printf("Send SIGUSR1 to %d\n", ::getpid());

	auto & instance = Singleton::getInstance();
	puts_signal_safe("returned from getInstance"sv);
	if (count_construction > 1)
	{
		puts_signal_safe("constructed more than once!"sv);
	}
	if (!instance.is_initialized())
	{
		puts_signal_safe("instance not initialized!"sv);
	}
	return 0;
}
```

mysterymath:
That's a fair point; interrupts can be used to build concurrent execution
(and indeed they're the only way to do so on a single processor system),
but their intrinsic behavior operates more like an unusual function call.
Application developers are wholly within their rights to use interrupt
handlers to swap out registers and the stack pointer on a timer; at that
point, you've got bonafide concurrent execution, since the two control flow
paths gain a degree of independence. I don't think we'd want to be too
prescriptive about how this might be done, so leaving the guard
implementations open for replacement would allow whatever implementation
makes the most sense.

common/lib/new.cc
karunski: How practical would it be to change the ABI expected by `clang++ --target=mos`, so that the guard variable doesn't take 8 bytes?  We only utilize 1.  Likewise, no need to return an int (uint8 would be sufficient).

karunski:
Ha; sure-- if someone comes up with a pre-emptive multitasking host that LLVM-MOS can target, I'll be first in line to make a thread-safe implementation here.

common/lib/new.cc
mysterymath: No idea; we'd have to look at the implementation in clang and see if it's amenable to change. Clang is surprisingly simple though; I'd be surprised if it's super hard.

karunski:
Rebased for the new build changes.

mysterymath (changes requested):

mos-platform/c64/lib/basic_header.S
```diff
@@ -1,11 +1,12 @@
...
 .section .basic_header,"aR",@progbits
 	.short next_line	; address of next BASIC line
 	.short 7773 		; line number: l33t for LLVM, a hint that this program was
 						; compiled with LLVM
 	.byte 0x9e			; SYS keyword BASIC token
-	.asciz "2061"		; the location of the _start symbol, in ASCII
+	.asciz "2062"		; the location of the _start symbol, in ASCII
 next_line:
 	.short 0   			; end of basic program
+	.byte 0             ; align to an odd number of bytes so that _start is at 2062 (2-byte aligned)
```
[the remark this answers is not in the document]
karunski: No; at least not to my knowledge; the issue is that .text is already over-aligned to 2 bytes, generating the pad byte implicitly. For what it's worth I only see it when compiling Debug, without optimizations.

The basic header is already aligning at 1 byte, as required for it to start at byte 0x801.

common/include/new
```diff
@@ -0,0 +1,29 @@
...
+namespace std {
+struct nothrow_t {
+  explicit nothrow_t() = default;
+};
+
+extern const nothrow_t nothrow;
+
+typedef void (*new_handler)();
+
+new_handler get_new_handler() noexcept;
```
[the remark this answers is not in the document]
karunski: I'm conflicted about the value of SJLJ exception handlng.  That design has a runtime cost and gave a lot of programmers a bad impression about exception support. SJLJ is arguably worse than simply not having exception support.  The modern exception implementation, using out-of-line exception tables, is not considered to be expensive at all. I have no idea why it conflicts with ELF support.  Doesn't the Linux-ELF target support exceptions?

I do assume that most users of the MOS target won't have much use for exceptions anyway; so in general I would call it a low-priority feature, but one that is ultimately required if you're going to try to get anywhere near standard compliance. It's really baked into the language at this point, and you have to write some pretty non-standard code to really avoid all the situations where a clang (or any modern implementation) will generate a throw.  For MOS; it's probably most important to just have predictable behavior in the exception-throwing situations: basically a good std::terminate() to jump to since it's unlikely that you'll ever have a good place to catch an exception. 

FWIW, compiling with -fno-exceptions absolutely does not mean your program won't generate an exception-- you'll still see exceptions thrown from within libc++ and __cxa_* functions; -fno-exceptions does nothing to change that. 

I don't know how much it matters on `std::get_new_handler`, but in general having the correct declaration of `noexcept` is increasingly important.  It's really an orthogonal issue from whether or not the code generator actually emits exception frames.  

Whether or not a function is `noexcept` is part of it's function signature, and as such it's something that plays a part in type selection for templates. See the `std::is_nothrow_*`  type_traits for examples. 

What's more I've already run into issues with the code clang++ generates for `new (std::nothrow) T{}` for type T that has a non-trivial constructor: clang++ will only check the return type from the underlying `operator new` implementation if  the `operator new(..., nothrow)` overload is marked `noexcept`.  Otherwise, it will blindly call the object constructor with `this` set to nullptr if the allocation fails!  Based on that experience, I am assuming that clang++ really does require functions implicitly defined in the standard library to correctly define non-throwing functions; otherwise bad things will occur.

mos-platform/c64/lib/basic_header.S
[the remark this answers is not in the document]
karunski: On this point-- I tried explicitly relaxing the alignment of the symbols that requested 2-byte alignement, but clang wouldn't do it.  It says minimum alignment is 2.

common/include/new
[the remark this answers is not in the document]
mysterymath: >The modern exception implementation, using out-of-line exception tables, is not considered to be expensive at all.
Very much depends on who you talk to. It's certainly 

> I have no idea why it conflicts with ELF support. Doesn't the Linux-ELF target support exceptions?
Ah, I misspoke; not ELF, but the CFI tables. It's very unlikely that the existing exception table generation code in LLVM is a practical implementation on the 6502, so it'd be a small researchey project to make a reasonable new one.

> I do assume that most users of the MOS target won't have much use for exceptions anyway; so in general I would call it a low-priority feature, but one that is ultimately required if you're going to try to get anywhere near standard compliance.

"C++11 standards compliance, except enforcing -fno-exceptions," is still a reasonable guarantee to offer to users; akin to the "C99 standards compliance, except without floats" we currently offer. We wouldn't be alone in that either: the NVPTX and AMDCPN GPU backends both operate exactly this way, to the point of simply overriding the -fexception flag immediately after it's parsed. There's an if statement in Clang there with our name on it if we want to go that route.

I do eventually want floats, just for completeness sake, but adding floats to the runtime is a considerably less invasive change than exceptions.

> FWIW, compiling with -fno-exceptions absolutely does not mean your program won't generate an exception-- you'll still see exceptions thrown from within libc++ and __cxa_* functions; -fno-exceptions does nothing to change that.

Sure it does; we're the ones writing libc++ and all the __cxa_ functions.
 
> What's more I've already run into issues with the code clang++ generates for `new (std::nothrow) T{}` for type T that has a non-trivial constructor: clang++ will only check the return type from the underlying `operator new` implementation if the `operator new(..., nothrow)` overload is marked `noexcept`. Otherwise, it will blindly call the object constructor with `this` set to nullptr if the allocation fails! Based on that experience, I am assuming that clang++ really does require functions implicitly defined in the standard library to correctly define non-throwing functions; otherwise bad things will occur.

In this case, I'd expect a `-fno-exceptions`-ey implementation of the "throwing" `new` to call `std::terminate`; in that case, the offending `this` deallocation would never be reached. The same could be done for every mandated exception in the standard library.

mos-platform/c64/lib/basic_header.S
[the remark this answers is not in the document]
mysterymath: Would you mind posting steps to reproduce the issue here? I smell a bug, and I'd at least like to give it a quick look before we work around it in this fashion.

mysterymath (changes requested):

common/include/new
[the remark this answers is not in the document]
karunski: > Very much depends on who you talk to. It's definitely not as bad as its worst detractors think, but it's also very uncommon in, say, the game dev world. To be fair, I learned this from a local talk by a Sony LLVM dev trying to lower the overhead and convince his company that exceptions are sorta conditionally OK now.

I would say-- ask the upstream clang developers. Game dev is a special case in that exceptions have limited usefulness even when they are free. My perspective is for programs that have to actual deal with unforseen exception circumstances exceptions are "free" in the sense that non-throwing code paths generate identical code regardless of whether exceptions are on.

> There's an if statement in Clang there with our name on it if we want to go that route.

I would say that the MOS target is more general purpose than the GPU. As such you are more likely to have users that want exceptions for their more complex, non-game, programs.  Doesn't have to be a high priority feature though.

> In this case, I'd expect a -fno-exceptions-ey implementation of the "throwing" new to call std::terminate; in that case, the offending this deallocation would never be reached. The same could be done for every mandated exception in the standard library.

I think you're missing the point on this one:  `operator new(..., nothrow)` is required to _not_ throw. If allocation fails, clang must skip the constructor call and return null. Clang++ only generates this code if your implementation of the operator function is marked noexcept.

mos-platform/c64/lib/basic_header.S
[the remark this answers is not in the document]
karunski: Sure; compile a program containing member functions, and instruct the linker to generate a map file.  You can see this in hello-new or hello-malloc.

When using the "Debug" cmake preset, the .text symbols in the map file will have 2-byte alignment.  The .text segment will specify 2 byte alignment.

When using "MinSizeRel", there are no .text symbols with 2-byte alignment.  The .text section has 1-byte alignment.

common/include/new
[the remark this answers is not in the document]
mysterymath: So, I had a quick discussion with the other project members about this, and we've reached a consensus.

While C++ exceptions do have their uses, they produce an unavoidable size overhead due to CFI tables whether or not exceptions are thrown, and are considerably slower than regular control flow when exceptions are thrown. Worse, exceptions are a leaky abstraction from a compiler author's POV; the possibility of an exception travelling through a code region on unwind requires special casing throughout the compiler. Accordingly, there's a considerable maintenance burden involved with supporting exceptions: in the runtime libraries, in the standard library, and in the compiler itself.

However, none of the above is a reason to permanently ban exception handling; just to put it off. If and when we get around to implementing proper exception handling, we should do so all in one sweep, from the compiler, through the  runtime libraries, to the standard library. This will help us better understand the maintenance costs involved and allow us to avoid paying any of these costs until then.

Accordingly, as soon as I get the chance, I'll be adding the necessary lines to force `-fno-exceptions` for the MOS target in clang.

The implications for this PR:
1. The `noexcept` specificiers for internal functions won't do anything. Please remove them; we can add them later.
2. Please also implement the "throw" versions of new as well; we can implement them up to the `-fno-excpetions` guarantee by having them call `std::terminate` in case of allocation failure.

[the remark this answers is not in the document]
karunski: I'm o.k. with running no-exceptions.  I'd rather we just leave them un-implemented so that code that tries to call throw just doesn't compile, which is the current status quo.

However:
> The noexcept specificiers for internal functions won't do anything. Please remove them; we can add them later.

This is demonstrably false.  The clang++ front-end generates different, incorrect code, if the noexcept is missing on the functions that require it.  Try it yourself if you don't believe me-- remove noexcept from the operator new() overloads and build/run the hello-new example.  Whether or not you compile with fno-exceptions has zero effect on this.

>Please also implement the "throw" versions of new as well; we can implement them up to the -fno-excpetions guarantee by having them call std::terminate in case of allocation failure.

These are already implemented at, for example new.cc line 300. At this point it calls abort() rather than std::terminate.  I can implement std::terminate in a subsequent PR and will swap it over at that point.

mos-platform/c64/lib/basic_header.S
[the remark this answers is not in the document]
mysterymath: Ah, I see the problem; because we hardcode the start address using ASCII, we can't allow the real start address of `.text` to vary with its alignment requirements. I'll make a separate bug for the issue; no action required here. Should be easy to fix, but some thought will be needed for the cleanest way.

common/include/new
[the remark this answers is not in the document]
mysterymath: By `internal functions`, I mean functions internal to the new.cc file; functions like `get_block()`.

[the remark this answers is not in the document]
mysterymath: > These are already implemented at, for example new.cc line 300. At this point it calls abort() rather than std::terminate. I can implement std::terminate in a subsequent PR and will swap it over at that point.

Sorry, I realized this immediately after posting. No worries; we can get to std::terminate later.

common/lib/new.cc
mysterymath: Created llvm-mos/llvm-mos#137.

mos-platform/c64/lib/basic_header.S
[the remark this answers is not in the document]
mysterymath: Created #14 for this issue.

mos-platform/common/lib/new.cc
```diff
@@ -0,0 +1,410 @@
...
+    // Standard c++ requires throwing bad_alloc here.
+    // TODO: implement throw.
+    // Calling abort() here behaves similarly to a caller that
+    // does not catch bad_alloc, and therefore has an unhandled exception.
+    abort();
+  }
+}
+} // namespace
+
+std::size_t blocklist::m_heap_limit = SIZE_MAX;
```
mysterymath, line 277: Add a comment noting that SIZE_MAX is a special-cased sentinel value, along with a summary of its semantics.

mos-platform/common/lib/new.cc
```diff
@@ -0,0 +1,410 @@
...
+
+void __set_heap_limit(size_t new_size) {
+
+  if (blocklist::m_heap_limit == SIZE_MAX) {
+    // Heap is uninitialized... set it up for the first call
+    blocklist::m_heap_limit = (new_size < blocklist::MIN_ALLOC_SIZE)
+                                  ? blocklist::MIN_ALLOC_SIZE
+                                  : new_size;
+  } else {
+    // heap is initialized.
```
mysterymath, line 402: nit: "Heap"

mos-platform/common/lib/new.cc
```diff
@@ -0,0 +1,410 @@
...
+
+    const auto freed_block_ptr = block_node::get_node(*block);
+    last_block_node->m_next = freed_block_ptr;
+    freed_block_ptr->m_prev = last_block_node;
+    freed_block_ptr->m_next = nullptr;
+    m_free += block->m_size;
+    m_used -= block->m_size;
+  }
+
+  void coallesce_blocks() noexcept {
```
mysterymath, line 191: nit: "coalesce"

mysterymath (changes requested):

mos-platform/common/lib/new.cc
```diff
@@ -0,0 +1,410 @@
...
+    for (auto &block : *this) {
+      if (block.m_size >= (sz + sizeof(block_node))) {
+        return &block;
+      }
+    }
+
+    return nullptr;
+  }
+
+  void free_block(block *const block) noexcept {
```
mysterymath, line 157: Add a TODO  comment for each of these operations that are O(n). This PR is for a simple heap implementation, but simple is a very relative term, and it's time complexity is an order of magnitude worse than typical, even for far simpler approaches than a full dlmalloc.

mysterymath (changes requested):

mos-platform/common/lib/new.cc
karunski: Commentary added. If someone comes up with an implementation for MOS that is really an order of magnitude faster in terms of real run time, I welcome them to show it.  

Keep in mind that the (n) in these operations is low: in practical terms it's in the 100's at most, and absolute maximum (n) is only 4096.

mysterymath:
Alright, I've gone through the remainder of new.cc; looks alright to go in, pending the remaining outstanding comments.

Thanks for taking the time to put this together, and for running it through the code review gauntlet.

mos-platform/c64/lib/stdlib.c
```diff
@@ -0,0 +1,12 @@
+// place holder for stdlib "abort" function.
+#include <stdio.h>
+
+// Not sure what the best way to handle this is on C64.
+
+void abort() {
```
mysterymath, line 6: A cross-platform abort has already been checked in; you'll likely need to sync.

mysterymath (changes requested):

mysterymath (approved):
Alright, LGTM. Feel free to merge at your convenience.

karunski:
Thanks; but I don't have write access on this repo.  Someone with write access will have to merge it.

PR #13, opened 2022-01-09
implement exit and abort for a800xl / c64

mrk-its:
It is required for https://github.com/llvm-mos/llvm-mos/issues/132


mysterymath:
So, this returns to BASIC, which I think is generally the right move for an exit implementation, but I'm not 100% confident that this will return to a fully-working BASIC. IIRC, our zero page variable selection just uses the first 16 zero page locations as registers, which BASIC uses for various purposes.

We'd need to be a bit more careful and only select locations that are used for temporary purposes in BASIC; that is, things that only hold meaningful values while non-SYS operations are being performed.

That may already be the case; I don't know affirmatively that any problems are presently possible; it's just that no determination was ever made one way or the other.

So, we can either:
1. Take the easy way out an make exit an infinite loop. I think that I recommended #11 do this (they also implemented an exit) but it's not really ideal.
2. Vet the zero page registers to find ones that we can safely clobber and return to BASIC.
3. Find zero page registers that we can clobber so long as we save and restore them around main.

My preferences run 2->3->1, since the C64 really does have something like an OS in the form of BASIC, and it'd be nice to play nice in that environment wherever possible.

mrk-its:
For C64 I'm preserving current behavior now (infinite loop): https://github.com/llvm-mos/llvm-mos-sdk/pull/13/files#diff-bdaebe5ff88d130810fc84fa7653049d9200d63be5477a25387c87db650a955eR13 - I would like to stay with it for now as I don't know C64 platform. We can fix it later if required.

common/lib/exit.c
```diff
@@ -0,0 +1,13 @@
+#include <stdlib.h>
+#include <stdio.h>
+
+extern void __exit(int status);
+
+void exit(int status) {
+  __exit(status);
+
+  // Prevent the compiler from considering this entire basic block unreachable.
+  asm volatile ("");
```
mysterymath, line 10: Instead, just mark __exit as attribute((noreturn)). This is done automatically for abort() and exit()

common/lib/main.s
```diff
@@ -0,0 +1,20 @@
+.text
+.global __main
+.global __exit
+
+__main:
+    TSX
+    STX __save_sp
+    JMP main
+
+__exit:
```
mysterymath, line 10: Add a comment describing the mechanics of how this works; in particular the logic of how it uses to the values saved in the `__main` wrapper.

common/lib/main.s
```diff
@@ -0,0 +1,20 @@
...
+   STX __save_x
```
mysterymath, line 11: `STX mos8(__rc2)`. Since __exit is a normal C function, the scratch zero page registers are available.

common/lib/main.s
```diff
@@ -0,0 +1,20 @@
...
+   LDX __save_sp
+   TXS
+   LDX __save_x
+   rts
+
+__save_sp:
```
mysterymath, line 17: __save_s; we've taking to calling the hardware stack register S and the soft stack pointer SP.

common/lib/main.s
```diff
@@ -0,0 +1,20 @@
...
+    TSX
+    STX __save_sp
+    JMP main
+
+__exit:
+   STX __save_x
+   LDX __save_sp
+   TXS
+   LDX __save_x
+   rts
```
mysterymath, line 15: nit: capitalize RTS

mysterymath (changes requested):

mos-platform/common/lib/exit.c
```diff
@@ -0,0 +1,8 @@
+#include <stdlib.h>
+#include <stdio.h>
+
+extern __attribute__((noreturn)) void __exit(int status);
```
mrk-its, line 4: I've put this attribute here, will it work?

mysterymath: Yep!

mos-platform/c64/lib/abort.c
```diff
@@ -0,0 +1,7 @@
+#include <stdlib.h>
+#include <stdio.h>
+
+void abort(void) {
+   puts("Aborted");
```
mysterymath, line 5: ABORTED; the C64 doesn't have lowercase letters by default.

mysterymath (approved):

PR #16, opened 2022-01-11
store s reg in .noinit section

mrk-its:


mos-platform/common/lib/main.S
```diff
@@ -20,5 +20,6 @@ __exit:
    LDX mos8(__rc2)
    RTS
 
+.data
```
mysterymath, line 23: Instead, use .noinit; this saves it from being included in the binary or memset to zero on init.

mysterymath (approved):

mos-platform/common/lib/main.S
mysterymath, line 23: Instead, use .noinit; this saves it from being included in the binary or memset to zero on init.

mysterymath (changes requested):

PR #17, opened 2022-01-15
Implement std::terminate

karunski:
Invoke terminate from within the new_handler instead of invoking abort() directly.


mos-platform/common/lib/exception.cc
```diff
@@ -0,0 +1,24 @@
+#include <exception>
+#include <stdlib.h>
+
+namespace std {
+    // __current_terminate_handler shall be never-null.
+    // It's either abort() or the function provided at set_terminate.
+    static terminate_handler __current_terminate_handler = ::abort;
+
+    // Current standard explicilty says the default handler value does not have to be nullptr.
+    terminate_handler get_terminate() noexcept {
```
mysterymath, line 10: Add a note that these aren't thread-safe in the presence of user thread implementations.

mysterymath (changes requested):

mos-platform/common/lib/exception.cc
mysterymath: Actually, if you would, do this in the header, not the implementation. More likely to be seen that way.

mysterymath: I've created llvm-mos/llvm-mos#138 as a potentially more principled solution for atomicity and thread safety in llvm-mos.

mysterymath: Actually, forget the whole thing. We should have this as another callout of things not yet supported by the compiler/SDK/runtime, like exceptions. No action required.

mysterymath (approved):

PR #18, opened 2022-01-15
Make std::set/get_new_handler work per standard.

karunski:
Doing so revealed some issues  in __set_heap_limit:
1. New memory needs to be coalesced into the previous blocks, so a large allocation spanning new and old heap limits can succeed.
2. The __heap_bytes_free accounting was not correct after extending the heap limit.

### DIFF
--- a/mos-platform/common/lib/new.cc
+++ b/mos-platform/common/lib/new.cc
@@ -259,7 +259,7 @@ public:
 
 private:
   // Nodes are added at the end of the list.
-  auto add_node(block_node *node) {
+  block_node * add_node(block_node *node) {
     list_node *node_last = m_head;
     for (; node_last->m_next; node_last = node_last->m_next) {
     }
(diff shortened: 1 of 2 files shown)


karunski:
To elaborate-- the std::set_new_handler could actually be useful for 6502 programs. By hooking into this function, the user is notified when their heap is full. The caller could then do useful actions, like moving data to a REU, or moving it to another bank, or extending the heap limit. The new operator will then retry the allocation if they succeed in making more memory free.

The associated unit test PR shows how to use this with __set_heap_limit.

mos-platform/common/lib/new.cc
```diff
@@ -330,24 +335,58 @@ __attribute__((weak)) void free(void *ptr) {
 
 }
 
+// All operator new implementations forward to this overload.
```
mysterymath, line 338: Move this to the header, since it impacts standards compliance.

mos-platform/common/lib/new.cc
```diff
@@ -243,31 +243,36 @@ class blocklist {
...
+      // Coalesce blocks so the new memory joins with the last free block,
+      // which is likely to be at the end of previous heap limit.
+      coalesce_blocks();
     }
   }
 
 private:
   // Nodes are added at the end of the list.
-  void add_node(block_node *node) {
+  auto add_node(block_node *node) {
```
mysterymath, line 262: Explicitly give the return type; it's not obvious from the function name.

mysterymath (changes requested):

mysterymath (approved):

PR #19, opened 2022-01-16
Initialization and finalization

karunski:
Invoke .init_array section generated by the compiler.

Implement __cxa_atexit, which is used by the compiler for both global and local static destructors.

Implement atexit, which forwards to __cxa_atexit.

Hook the invocation of __cxa_atexit registrations into the exit() function.


mos-platform/common/lib/exit.c
```diff
@@ -2,7 +2,9 @@
 #include <stdio.h>
 
 extern __attribute__((noreturn)) void __exit(int status);
+extern void __fini();
 
 void exit(int status) {
+  __fini();
```
mysterymath, line 8: `abort()` currently calls through to `exit`, so as written, this will cause `abort()` to call functions registered with `atexit()`, in violation of the standard.

mos-platform/common/lib/cxa_abi.cc
```diff
@@ -0,0 +1,157 @@
...
+// Static allocation of registration list.
+RegistrationList::FnBlock RegistrationList::m_tail{};
+RegistrationList::FnBlock * RegistrationList::m_list = &m_tail;
+}
+
+// Currently the compiler does not generate any calls to this __cxa_finalize.  Finalization
+// is invoked from __fini which is called from exit();
+static void __finalize_noargs() { RegistrationList::run_all_exits(); }
+
+// atexit / finalize are implemented under the assumptin that there is only a single
```
mysterymath, line 144: nit: `assumption`

mos-platform/common/lib/cxa_abi.cc
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    static void run_all_exits() {
+        while (m_list != &m_tail) {
+            m_list->run_all();
+
+            const auto current_node_ptr = static_cast<FnNode *>(m_list);
+            const auto next_block = current_node_ptr->m_next;
+            delete current_node_ptr;
```
mysterymath, line 119: No need; we're about to exit the program. At least until we have DSO unloading, which is an indefinite amount of time in the future.

mysterymath (changes requested):

mos-platform/common/ldscripts/c_sections.ld
```diff
@@ -4,7 +4,29 @@ __load_begin = .;
 .text : { *(.start) *(.text .text.*) }
 .rodata : { *(.rodata .rodata.*) }
 .data : { *(.data .data.*) }
+
+/* .init_array is a sorted list of initialization function pointers.
```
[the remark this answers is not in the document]
karunski: The CxaFinalizer entry (in .fini_array), is modified when cxa_atexit is first invoked.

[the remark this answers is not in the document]
mysterymath: Ah, just make a note of that here then.

mos-platform/common/lib/exit.c
karunski: Unless I'm missing something, this is a bug isn't it?  Abort() shouldn't call through exit(); it should terminate immediately.

mysterymath: It's a bug with this PR, but it wasn't before. Previously, exit had no behavior besides terminating.

karunski: OK; so I should modify abort() so it goes straight to __exit()?

mysterymath: Yeah, SGTM

mos-platform/common/lib/main.S
```diff
@@ -8,7 +8,7 @@
 __main:
     TSX
```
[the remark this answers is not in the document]
karunski: I need help with this one.  My meager assembly language skills are failing me. If it make __save_s a routine called from main(), how do I get the correct stack entry to store in __exit_s?  If I just take the existing stack value I think I end up back in __main, rather than the caller for __main?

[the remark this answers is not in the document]
mysterymath: Ah, you're right; I hadn't thought this through completely; it'd need to be something that could be injected directly inline, which is more complex than what you already have.

No action required.

mysterymath:
LGTM, thanks for the PR!

Issue #14, opened 2022-01-11
Alignment of C64 text section not respected


mysterymath:
Because the position of the _start symbol is hard coded in ASCII in the BASIC header, if the start symbol moves due to text section alignment, the BASIC header will jump to the wrong location. The implementation should be made tolerant to any alignment of the text section.

PR #21, opened 2022-01-17
Use `.mos_addr_asciz` directive to relocate `_start` for BASIC.

jackoalan:


mysterymath (approved):
Awesome, LGTM!

mysterymath:
Fixes #14.

johnwbyrd:
Note that basic_header.s could be promoted to a common directory now.  Most 8-bit micros, including but not limited to Commodore 128, Vic 20, and Apple IIe, all share this new header format.

Issue #22, opened 2022-01-18
Implement realloc


mrk-its:
It would be great to have also `realloc` implemented, especially for small sizes of heap.
For example, I'm testing allocation using following rust code:

```
input.split("\n").map(|i| i.parse::<i16>().unwrap()).collect::<Vec<_>>();
```

and it fails trying to allocate memory for 2000 integers (4000 bytes total) having default heap size 4096 bytes.

It produces following chain of calls of rust allocator:

```
alloc(2) -> 0x520e
alloc(8) -> 0x5216
realloc(0x520e, 8) -> 0x5216
alloc(16) -> 0x5224
realloc(0x5216, 16) -> 0x5224
alloc(32) -> 0x523a
realloc(0x5224, 32) -> 0x523a
alloc(64) -> 0x5260
realloc(0x523a, 64) -> 0x5260
alloc(128) -> 0x52a6
realloc(0x5260, 128) -> 0x52a6
alloc(256) -> 0x532c
realloc(0x52a6, 256) -> 0x532c
alloc(512) -> 0x5432
realloc(0x532c, 512) -> 0x5432
alloc(1024) -> 0x5638
realloc(0x5432, 1024) -> 0x5638
alloc(2048) -> 0x0
realloc(0x5638, 2048) -> 0x0
PANIC!!!
```

So it starts with small memory allocated for vector and increases capacity as new data is added trying to realloc existing buffer. Default realloc implementation simply allocates new buffer and copies data - that's why there is not enough memory.

relloc would reduce heap fragmentation in such case and simply fit data in memory.

cc: @karunski

karunski:
I think this is possible. You can assign to me.

karunski:
@mrk-its  Please try the realloc implementation. It will realloc in the manner you describe.

mrk-its:
@karunski thanks! will do that later today and let you know.

mrk-its:
Works like a charm!:
```
heap bytes free: 4090 / 4096
alloc 2 -> 0x5998
realloc 0x5998 8 -> 0x5998
realloc 0x5998 16 -> 0x5998
realloc 0x5998 32 -> 0x5998
realloc 0x5998 64 -> 0x5998
realloc 0x5998 128 -> 0x5998
realloc 0x5998 256 -> 0x5998
realloc 0x5998 512 -> 0x5998
realloc 0x5998 1024 -> 0x5998
realloc 0x5998 2048 -> 0x5998
[...]

PR #23, opened 2022-01-22
Realloc

karunski:
This implementation of realloc handles the scenario of "growing an existing alloc into free space", which is described in #22

### DIFF
--- a/mos-platform/common/include/string.h
+++ b/mos-platform/common/include/string.h
@@ -6,6 +6,9 @@
 
 #ifdef __cplusplus
 extern "C" {
+#define __RESTRICT
+#else
+#define __RESTRICT restrict
 #endif
 
 void *memchr(const void *s, int c, size_t n);
@@ -18,11 +21,11 @@ char *strchr(const char *s, int c);
 
 int strcmp(const char *s1, const char *s2);
 
-char* strcpy(char *restrict s1, const char *restrict s2);
+char *strcpy(char * __RESTRICT s1, const char * __RESTRICT s2);
 
 size_t strlen(const char *s);
 int strncmp(const char *s1, const char *s2, size_t n);
-char *strncpy(char *restrict s1, const char *restrict s2, size_t n);
+char *strncpy(char * __RESTRICT s1, const char * __RESTRICT s2, size_t n);
 
 char *strrchr(const char *s, int c);
 
--- a/mos-platform/common/include/string.h
+++ b/mos-platform/common/include/string.h
@@ -6,9 +6,6 @@
 
 #ifdef __cplusplus
 extern "C" {
-#define __RESTRICT
-#else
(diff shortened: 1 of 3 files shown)


mos-platform/common/include/string.h
```diff
@@ -18,11 +21,11 @@ char *strchr(const char *s, int c);
 
 int strcmp(const char *s1, const char *s2);
 
-char* strcpy(char *restrict s1, const char *restrict s2);
+char *strcpy(char * __RESTRICT s1, const char * __RESTRICT s2);
```
mysterymath, line 24: Try `__restrict__`; this should work in Clang for both C and C++.

mos-platform/common/lib/new.cc
```diff
@@ -293,6 +335,20 @@ blocklist &get_free_list() {
   return free_list;
 }
 
+// Implement a reallocation by allocating a new region and copying
+// old data to new.
+void * realloc_copy(void * orig, size_t sz) {
+  const auto new_alloc = malloc(sz);
+  if (new_alloc) {
```
mysterymath, line 342:
```
if (!new_alloc)
  return nullptr;
```
Saves the indentation.

mos-platform/common/lib/new.cc
```diff
@@ -311,6 +367,35 @@ __attribute__((weak)) void *malloc(size_t count) {
   return free_list.split_block(free_list.find_first_fit(count), count);
 }
 
+__attribute__((weak)) void *realloc(void *orig, size_t count) {
+  if (!orig) {
+    return malloc(count);
+  }
+
+  const auto orig_block_ptr = block::get_block(static_cast<std::byte *>(orig));
+  if (count > orig_block_ptr->m_size) {
```
mysterymath, line 376:
```
if (count <= orig_block_ptr->m_size) {
  // No realloc ....  
  return orig;
}
```
Saves the indentation.

mos-platform/common/lib/new.cc
```diff
@@ -311,6 +367,35 @@ __attribute__((weak)) void *malloc(size_t count) {
...
+    auto &free_list = get_free_list();
+    const auto grow_sz = count - orig_block_ptr->m_size;
+
+    // Find the adjacent block, and if found split it.
+    const auto adjacent_block =
+        free_list.find_adjacent_fit(orig_block_ptr, grow_sz);
+    if (adjacent_block) {
```
mysterymath, line 383: Similarly, `if (!adjacent_block) ... return`, then the remainder can be dedented.

mysterymath (changes requested):

mos-platform/common/include/string.h
karunski: Good to know! Thank you!

mysterymath:
LGTM!

PR #28, opened 2022-01-31
RTTI and dynamic_cast from libcxxabi

karunski:
This deserves some explanation-- this is the implementation of ABI-required type_info and dynamic_cast based on the libcxxabi. 

The libcxxabi implementation does work as is, unmodified, but in unmodified form had the effect of including a big chunk of the exception handling implementation (can_catch) and all of the dynamic_cast algorithm, even if the program being compiled did not make use of those features. This is due to extending the virtual function tables of type_info derived classes to include this functionality; and since the vtables reference the implementations; the linker cannot optimize them away when they aren't used.   All told, it added about 8k of text to any program that had any vtable usage; which is completely inconsequential for a modern clang targer, but seems like a big deal for an 8-bit target.

I have changed the implementation as follows:
* Remove the exception handling methods; at least until we have some way to use them. Saving about 2k of compiled code.
* Resize parameters and some data fields from int to unsigned char, saving about 500 bytes in the binary.
* Refactor dynamic_cast related functions to non-vtable calls; so they are only reachable from within the dynamic_cast implementation.  This successfully removes any cost of dynamic_cast unless the program uses it, and still allows simple usage of typeid() / std::type_info without paying for the full dynamic_cast impl.  

This implementation successfully runs the dynamic_cast_algorithm tests from llvm-test-suite/ABI-Testsuite/test/s2_9/dynamic_cast_algorithm.xpp.  However; making the whole ABI-Testsuite work still looks like a huge job-- it seems to assume 32-bit or 64-bit pointers only.


mos-platform/common/lib/__cxxabi_config.h
```diff
@@ -0,0 +1,100 @@
...
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef ____CXXABI_CONFIG_H
+#define ____CXXABI_CONFIG_H
+
+#if defined(__arm__) && !defined(__USING_SJLJ_EXCEPTIONS__) &&                 \
```
mysterymath, line 12: Can we clean this up a bit, take out the stuff that "obviously" doesn't apply? That would make it quite a bit easier to review, since I'm going to have to learn how this algorithm works.

karunski: OK; I removed the sections that didn't look like they'd ever be compiled in.

mos-platform/common/lib/private_typeinfo.cpp
```diff
@@ -0,0 +1,945 @@
...
+// above.
+// If it finds a dst_type node it should search base classes using search_above_dst
+// to find out if this dst_type points to (static_ptr, static_type) or not.
+// Either way, the dst_type is recorded as one of two "flavors":  one that does
+// or does not point to (static_ptr, static_type).
+// If this is neither a static_type nor a dst_type node, continue searching
+// base classes above.
+// All the hoopla surrounding the search code is doing nothing but looking for
+// excuses to stop the search prematurely (break out of the for-loop).  That is,
+// the algorithm below is simply an optimization of this:
```
karunski, line 367: The simplified form of the search algorithm listed in this comment would probably be noticeably smaller, at the expense of being slower.

mos-platform/common/lib/private_typeinfo.cpp
```diff
@@ -0,0 +1,945 @@
...
+
+struct search_funcs {
+  void (*search_above_dst)(const __class_type_info *, __dynamic_cast_info *,
+                           const void *, const void *, path_accessibility,
+                           bool);
+  void (*search_below_dst)(const __class_type_info *, __dynamic_cast_info *,
+                           const void *, path_accessibility, bool);
+};
+
+enum class dyn_class_type : unsigned char {
```
mysterymath, line 857: Add implementation comments describing why this is necessary.

mysterymath (changes requested):
Just to double-check, none of this will end up in the final binary if -fno-rtti is used, right?

mos-platform/common/lib/private_typeinfo.cpp
mysterymath: That's usually the tradeoff we'd want here, but it'd be nice to have the details on how much slower and how much smaller.

I also suspect that the commented-out code has probably bit-rotted at least a bit...

mysterymath: No action required; we can clean this sort of thing up at a later date.

mos-platform/common/lib/private_typeinfo.cpp
karunski: Correct; -fno-rtti means that none of the type_info instances are generated for any of the dynamic classes.

I've added the requested comments.

mysterymath (approved):

PR #30, opened 2022-02-12
Implement getchar

karunski:
Working great on C64 and sim... left Atari 8-bit stubbed out as I don't have any expertise there.


mos-platform/c64/ldscripts/kernal.ld
```diff
@@ -0,0 +1,39 @@
+PROVIDE(CINT   = 0XFF81);
```
mysterymath, line 1: Make sure to use reserved symbols for these; otherwise, we run afoul of C standards compliance. The standard library might incorrectly call user code (or worse, call a user _variable_) if they define one of these symbols.

mysterymath (changes requested):

mos-platform/c64/lib/getchar.c
```diff
@@ -0,0 +1,53 @@
...
+// I found online, and through my own testing, it's not safe to do other I/O
+// until the input is fully retrieved-- for example calling CHROUT after
+// fetching the first buffered character seems to cause the rest of the input
+// buffer to drop.  Thus, the entire line is held in our own input buffer.  
+// Theoretically this maps neatly to the buffer required for the stdin stream.
+
+char __getchar() {
+
+  // While the input buffer is not empty, return characters from it.
+  if (buf_begin != buf_end) {
```
mysterymath, line 29: There's a bit of duplication between the beginning and ending of this function, which suggests it could be refactored a bit.

How about:
```
char __getchar() {
  if (buf_begin == buf_end)
   __fill_buffer();
  return input_buffer[buf_begin++];
}

void __fill_buffer() {
  buf_begin = 0;
  ...
  for (;;) {
    ...
  }
}

mos-platform/c64/ldscripts/kernal.ld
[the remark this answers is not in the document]
karunski: What I was after here was a way to generate the call to kernal routine CHRIN such that clang/llvm would treat it as a normal function call, and thus generate `jsr __CHRIN`.  The only other thing I tried was the c-style solution of casting integer 0xFFCF to a function pointer of type `void(*)()`, If I did that, I always got an indirect call rather than a direct call.   Defining the routine as an explicit pre-defined symbol for the linker seemed to do the trick.  If there's a more elegant way of doing this I'm willing to learn.  I definitely don't know the best way of doing things in assembly.

mos-platform/c64/lib/getchar.c
karunski: Yes, your version is better.

mos-platform/c64/ldscripts/kernal.ld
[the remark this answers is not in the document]
mysterymath: I think we're mostly on thee same page if you put the code snippet I gave in a "kernal.s" file, and include it in the link, then it's exactly equivalent to the linker script snippet you're using in terms of the behavior of the C and assembly code. It just uses the ordinary way of defining symbols: including object files in the link. The linker script mechanism is more intended for post-link hackery.

PR #31, opened 2022-02-27
Type Traits Implementation

karunski:
This is an implementation of type_traits as of c++17. 

Most of the type traits map directly to clang intrinsics, so they have simple definitions.

The more complex template meta programs are simplified versions of the libc++ definitions. The primary differences are removing support for non-clang compiler and pre-c++17 variants.

### DIFF
--- a/mos-platform/common/include/__type_traits_impl.h
+++ b/mos-platform/common/include/__type_traits_impl.h
@@ -1,0 +1,41 @@
+#ifndef __TYPE_TRAITS_BASE_H__
+#define __TYPE_TRAITS_BASE_H__
+
+// This implementation header factors out common type traits
+// needed in multiple headers.
+// Some portions of type_traits depend on cstddef, and 
+// Some portions of cstddef depend on type_traits.
+
+namespace std
+{
+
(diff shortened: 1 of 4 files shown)


mos-platform/common/include/__type_traits_impl.h
```diff
@@ -0,0 +1,41 @@
+#ifndef __TYPE_TRAITS_BASE_H__
```
mysterymath, line 1: TYPE_TRAITS_IMPL_H__

mos-platform/common/include/__type_traits_impl.h
```diff
@@ -0,0 +1,41 @@
...
+#define __TYPE_TRAITS_BASE_H__
+
+// This implementation header factors out common type traits
+// needed in multiple headers.
+// Some portions of type_traits depend on cstddef, and 
+// Some portions of cstddef depend on type_traits.
+
+namespace std
```
mysterymath, line 9: nit: join on same line

mysterymath (approved):
LGTM

PR #32, opened 2022-03-03
Update fake6502.c

schlupa:
These macros are dangerous without parenthesis. 
`if(setsign() != 0)` would not do what people expect.

`if(status |= FLAG_SIGN != 0)` would evaluate to `if(status |= 1)` which is wrong.


jnutting:
💯

mysterymath (approved):

PR #37, opened 2022-05-14
Initial implementation of Ohio Scientific Challenger 1P target

smuehlst:
This is a first attempt to implement an llvm-mos target for the [Ohio Scientific](https://en.wikipedia.org/wiki/Ohio_Scientific) Challenger 1P computer. So far `__putchar()` has been implemented, and the printf()-based examples are working on emulators and on a real machine.

The `__putchar()` implementation was done as a C++ template class, with the intention that it can be overridden with other instantiations for different screen layouts. The real Challenger 1P uses a subset of the screen memory for a 25x25 character layout. On the other hand there exists for example the [Briel Superboard III](http://www.brielcomputers.com/wordpress/?cat=39) that is a modern Challenger 1P clone where the screen layout can be set to 32x32.


mos-platform/osi-c1p/abort.c
```diff
@@ -0,0 +1,8 @@
+#include <stdlib.h>
+#include <stdio.h>
+
+void abort(void) {
+   puts("ABORTED");
+   while (1);
```
mysterymath, line 6: It seems like the `while(1);` here is redundant, since exit-loop is used.

mos-platform/osi-c1p/osi_screen.h
```diff
@@ -0,0 +1,94 @@
...
+
+            // Fill last line with blanks
+            memset(last_line_start, ' ', screen_width);
+        }
+    }
+
+public:
+    static void clrscr(void)
+    {
+        memset((void *) scr_base, 0x20, video_ram_size);
```
mysterymath, line 61: ' ', for consistency with line 54

mos-platform/osi-c1p/osi_screen.h
```diff
@@ -0,0 +1,94 @@
...
+        cursor_x = cursor_y = 0;
+    }
+
+    static void cputc(char c)
+    {
+        if (c == '\n')
```
mysterymath, line 67: It looks like \n, ASCII LF (Line Feed), is being used to return the carriage, while \r, ASCII CR (Carriage Return), is being used to advance the line.

If this is is a specific OSC-ism, no action required, otherwise, definitely do reverse this.

mos-platform/osi-c1p/putchar.cc
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * @brief __putchar implementation for Challenger 1P
+ * 
+ * For other screen layouts this module must be overridden.
+ * 
+ * @param c 
+ */
+extern "C"
+void __putchar(char c)
```
mysterymath, line 27: See the C standard: https://port70.net/~nsz/c/c99/n1256.html#5.2.2

In particular, for C, the expectation for '\n' is that it should move to the *initial* position on the next line, while '\r' moves to the initial position on the *current* line. This is in conflict with the traditional use of CR and LF on printers and such, but much C code expects it.

Typcially, CRLF systems add a translation layer under the C standard interfaces (here, backed by `__putchar`) to emit the right sequence of underlying control codes. For your target, you could either translate here, or break from the usual conio convention in `cputc`. I think it'd probably be a good idea to eventually support a CC65-compatible conio, so translating seems preferable to me, but your call.

mos-platform/osi-c1p/osi_screen.h
smuehlst: Good catch! The output functions are based loosely on the conio assembler implementation for the osic1p target in cc65, and this bug was actually inherited from there.

mysterymath (changes requested):
Marking this as changes needed; clicked the wrong button last time.

smuehlst:
Implemented the requested changes.

mysterymath:
LGTM

PR #38, opened 2022-05-25
Clear screen via crt0.c

smuehlst:
- Fixed a bug in the osic1p routine to clear the screen.
- osic1p programs now clear the screen before entering `main`.


mos-platform/osi-c1p/crt0.c
```diff
@@ -0,0 +1,10 @@
+extern void __clrscr(void);
+
+asm(
+  ".section .init.40,\"axR\",@progbits\n"
+  "  jsr __late_init\n"
```
mysterymath, line 5: You might consider just `jsr __clrscr`, at least until there's something else in `__late_init`. I took a look at the NES target; we could probably do that there too (but no action required for this PR); if I recall correctly, there used to be more stuff in the NES `__late_init`.

In that case, you could just make this a .s file, which seems a bit cleaner than inline assembly.

mysterymath (approved):

smuehlst:
I amended the pull request with a commit that fixes #39.

mysterymath:
LGTM, but it looks like you need to pull/rebase first.

mos-platform/osi-c1p/crt0.c
smuehlst: I replaced the crt0.c module with a crt0.s module that calls ``__clrscr`` directly.

smuehlst:
> LGTM, but it looks like you need to pull/rebase first.

How would I do that? As the `osic1p` branch is already published to GitHub and was involved in a preceding pull request, is it actually possible to rebase it?

mysterymath:
You can do a merge instead if you like; I'm planning on squashing down the commits into one, so the specifics don't matter as much.

smuehlst:
> You can do a merge instead if you like; I'm planning on squashing down the commits into one, so the specifics don't matter as much.

I already  have incorporated the latest changes from `main` including commit 4565c07296a3654cc441b5457792048b79b0c4af into the `osic1p` branch.

mysterymath:
Sorry, my bad, brain fart.

PR #43, opened 2022-06-11
getchar() implementation for osi-c1p target

smuehlst:
This is the implementation of getchar() for the osi-c1p target. It is based on the cc65 implementation for reading the keyboard, which in turn was inspired by the original ROM routine of the Ohio Scientific Challenger 1P machine.

### DIFF
--- a/mos-platform/osi-c1p/getchar.c
+++ b/mos-platform/osi-c1p/getchar.c
@@ -1,13 +1,14 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+extern void __kbhit(void);
 extern unsigned char volatile __CHARBUF;
 
 int getchar(void)
 {
     do
     {
-        asm volatile ("JSR\t__kbhit" ::: "p", "a", "x", "y");
+        __kbhit();
     } while (!__CHARBUF);
     
     int const result = __CHARBUF;
--- a/mos-platform/osi-c1p/getchar.c
+++ b/mos-platform/osi-c1p/getchar.c
@@ -1,7 +1,18 @@
 #include <stdio.h>
 #include <stdlib.h>
 
-int getchar() {
-  puts("getchar() not yet implemented");
-  abort();
+extern unsigned char volatile __CHARBUF;
+
+int getchar(void)
+{
+    do
+    {
+        asm volatile ("JSR\t__kbhit" ::: "p", "a", "x", "y");
+    } while (!__CHARBUF);
+    
+    int const result = __CHARBUF;
+
+    __CHARBUF = 0;
+
+    return result;
 }
--- a/mos-platform/osi-c1p/link.ld
+++ b/mos-platform/osi-c1p/link.ld
@@ -2,12 +2,10 @@
  * OSI Challenger 1P Linker Script
  * RAM size 32 kB (TODO make this configurable)
  */
-
  
 /* 
  * Provide 16 imaginary (zero page) registers located in 0x02 - 0x21 range.
  */
-
 PROVIDE(__rc0 = 0x02);
 INCLUDE imag-regs.ld
 ASSERT(__rc0 == 0x02, "Inconsistent zero page map.")
@@ -15,16 +13,16 @@ ASSERT(__rc31 == 0x21, "Inconsistent zero page map.")
 
 MEMORY {
     ram (rw) : ORIGIN = 0x200, LENGTH = 0x7E00
-    zpstorage (rw): ORIGIN = __rc31 + 1, LENGTH = 0x20
+    zpstorage (rw): ORIGIN = __rc31 + 1, LENGTH = 0x05
 }
 
 SECTIONS {
-    .zp : { *(.lowmemory); } > zpstorage
+    .zpkbd : { *(.lowmemkbd); } > zpstorage
     INCLUDE c.ld
 }
 
 /* Set initial soft stack address to end of RAM area. (It grows down.) */
-/* TODO make dependent on RAM size */
+/* TODO make RAM size configurable */
 __stack = 0x7FFF;
 
 OUTPUT_FORMAT {
--- a/mos-platform/osi-c1p/link.ld
+++ b/mos-platform/osi-c1p/link.ld
@@ -17,7 +17,7 @@ MEMORY {
 }
 
 SECTIONS {
-    .zpkbd : { *(.lowmemkbd); } > zpstorage
+    .zp : { *(.zp) *(.zp.*)  } > zpstorage
     INCLUDE c.ld
 }
 
--- a/mos-platform/osi-c1p/link.ld
+++ b/mos-platform/osi-c1p/link.ld
@@ -3,14 +3,7 @@
  * RAM size 32 kB (TODO make this configurable)
  */
 
-MEMORY {
-    ram (rw) : ORIGIN = 0x200, LENGTH = 0x7E00
-}
-
-SECTIONS {
-    INCLUDE c.ld
-}
-
+ 
 /* 
  * Provide 16 imaginary (zero page) registers located in 0x02 - 0x21 range.
  */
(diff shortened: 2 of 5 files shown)


mos-platform/osi-c1p/link.ld
```diff
@@ -2,26 +2,27 @@
...
+ASSERT(__rc0 == 0x02, "Inconsistent zero page map.")
+ASSERT(__rc31 == 0x21, "Inconsistent zero page map.")
 
 MEMORY {
     ram (rw) : ORIGIN = 0x200, LENGTH = 0x7E00
+    zpstorage (rw): ORIGIN = __rc31 + 1, LENGTH = 0x05
 }
 
 SECTIONS {
+    .zpkbd : { *(.lowmemkbd); } > zpstorage
```
mysterymath, line 20: It's a bit more typical to have one output section per "kind of thing"; in this case that'd be ZP locations (i.e., `.zp : { ... }`. The linker can collate many input sections together from different files into one output section.

Similarly, the only reason to have more than one input section is to allow garbage-collecting unreferenced input sections. Even then, it's convention to have the input sections begin with the name of the output section (i.e., `.zp.kbd`). This allows the linker script to collect all such sections with a wildcard: `.zp : { *(.zp) *(.zp.*) }`. That way, there's no need to keep changing the linker script when additional ZP sections are added.

mysterymath (approved):

mos-platform/osi-c1p/link.ld
smuehlst: Ok, I can change that.

But what is the strategy to compute the size of the output region if more than one input section is assigned to it? At the moment I know that there are exactly five bytes consumed by the .lowmemkbd input section, so I'm writing that literally as the length of the zpstorage memory region.

mysterymath: Generally, you allocate as much space possible to the region in the linker script, and rely on the linker to place everything. So the zpstorage region would have all available zp locations, the linker would sum up the sizes of all the zp sections, and it they don't fit, it's a linker error. If they do fit, any unused space is left unitialized.

There's also the question of what to do if there's initialized data in the ZP sections; right now, most targets don't really have a way to handle that. Haven't spent too terribly much time thinking about what should happen there; maybe we'd want to split things into .zp.data and .zp.bss? No action required here; it's just something that's a bit underexplored at the moment.

mos-platform/osi-c1p/getchar.c
```diff
@@ -1,7 +1,18 @@
...
-int getchar() {
-  puts("getchar() not yet implemented");
-  abort();
+extern unsigned char volatile __CHARBUF;
+
+int getchar(void)
+{
+    do
+    {
+        asm volatile ("JSR\t__kbhit" ::: "p", "a", "x", "y");
```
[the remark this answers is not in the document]
smuehlst: If I understand you correctly writing this as a C call this would clobber imaginary registers, and this would generate additional save and restore code for these registers. If that is the case, the `asm` call does save some code.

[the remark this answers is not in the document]
mysterymath: Only if the compiler actually does decide to put anything in the imaginary registers for this; should be easy to check in the generated assembly.

mos-platform/osi-c1p/link.ld
smuehlst: > Generally, you allocate as much space possible to the region in the linker script, and rely on the linker to place everything. So the zpstorage region would have all available zp locations, the linker would sum up the sizes of all the zp sections, and it they don't fit, it's a linker error. If they do fit, any unused space is left unitialized.

But wouldn't that waste the unused space then?
 
> There's also the question of what to do if there's initialized data in the ZP sections; right now, most targets don't really have a way to handle that. Haven't spent too terribly much time thinking about what should happen there; maybe we'd want to split things into .zp.data and .zp.bss? No action required here; it's just something that's a bit underexplored at the moment.

I'm currently calling a function from crt0.s that initializes the data for the keyboard reading function, see

https://github.com/smuehlst/llvm-mos-sdk/blob/a136aaf17111d7569637ebc00950dc7f495887ae/mos-platform/osi-c1p/kbhit.s#L73-L81

mysterymath: > But wouldn't that waste the unused space then?
Why would it? It's just telling the linker "it's an error if you try to put ZP values past this point"; but nothing precludes you from monkey-hammering at the rest of the ZP in code if you know for sure that the memory is otherwise unused.
 
> I'm currently calling a function from crt0.s that initializes the data for the keyboard reading function, see
> 
> https://github.com/smuehlst/llvm-mos-sdk/blob/a136aaf17111d7569637ebc00950dc7f495887ae/mos-platform/osi-c1p/kbhit.s#L73-L81
You can use .db.b 1 and initialize the values manually, but you also could have written .byte <initial_value>, and the assembler would have put the value in the section, just like other code and data that the linker manages. Right now, that data would be ignored by the linker, since the OUTPUT_SECTION script doesn't say what to do with it. Ideally, it would be loaded into the zero page by the same process that loads the program into RAM to begin with, saving the extra step (and code) at initialization time.

smuehlst: I update the pull request with two commits to address the comments. The call to `__kbhit` via a C function call did not change the generated code at all.

> You can use .db.b 1 and initialize the values manually, but you also could have written .byte <initial_value>, and the assembler would have put the value in the section, just like other code and data that the linker manages. Right now, that data would be ignored by the linker, since the OUTPUT_SECTION script doesn't say what to do with it. Ideally, it would be loaded into the zero page by the same process that loads the program into RAM to begin with, saving the extra step (and code) at initialization time.

I don't fully understand this. Do you mean I actuall could get rid of the `__initkbhit` routine now if I would use `.byte` and if I would modify the linker script as necessary, or would this be a future option?

mysterymath: You might be able to do this now, depending on how exactly the code produced by the linker actually gets loaded into the OSI. It looks like your OUTPUT_FORMAT script starts at 0x200; would it work if it started at 0x00 instead, covering the zero page and stack too? If so, then I'd expect you could use .byte and have the data included into the system image; you'd just need to make another memory region, something like "ram_image", that covers the entirety of ram (memory regions are allowed to overlap), and use that in your OUTPUT_FORMAT instead.

mysterymath: If it doesn't work that way, then the situation is more complicated, and we could probably go over it outside of this code review. There's a way to have the data automatically memcpy-ed at init time; this is used e.g. when the initialization data is stored in ROM, but needs to be accessed in RAM, say for the NES target. The copy-data library stuff.

smuehlst: > You might be able to do this now, depending on how exactly the code produced by the linker actually gets loaded into the OSI. It looks like your OUTPUT_FORMAT script starts at 0x200; would it work if it started at 0x00 instead, covering the zero page and stack too? If so, then I'd expect you could use .byte and have the data included into the system image; you'd just need to make another memory region, something like "ram_image", that covers the entirety of ram (memory regions are allowed to overlap), and use that in your OUTPUT_FORMAT instead.

Ah, I see. I think that this is way over my head linker-script-wise at the moment :-)

I will keep that in mind for a future improvement, especially because there's another feature present in cc65 that I would like to port eventually. Currently the whole binary has to be converted with the `srec_cat` program into a special hex format that can be loaded over the serial port. The cc65 osic1p target has the option to generate a hybrid format where only a small boot loader program is encoded in the hex format, and the actual program is then appended as binary data, which improves the load time drastically. See https://github.com/cc65/cc65/blob/master/libsrc/osic1p/bootstrap.s.

So let's leave it at that for the moment if that is ok with you.

mysterymath: Sounds good to me!

Issue #44, opened 2022-06-21
MEGA65 target


mlund:
I have set up a [small project](https://github.com/mlund/mega65-adventure) to use llvm-mos-sdk with rust with the aim to run on the [Mega65](https://www.mega65.org) computer. This has a C64 compatibility mode (`go 64` a la C128) where the test program runs fine. I wonder what is required to make a true Mega65 target? I'm aware that the mos-llvm project currently doesn't generate CSG4510 instructions (based on 65CE02), but that's perhaps not initially a problem as it's 6502 compatible. I have started a [fork](https://github.com/mlund/llvm-mos-sdk) with what I think are the structural parts, but so far it's merely a clone of the C64 target. Any input would be highly appreciated.

mysterymath:
Hey Mikael,

Thanks for reaching out! We have two guides on our Wiki that might be
relevant:
First, the Porting <https://llvm-mos.org/wiki/Porting> guide, which
describes how the compiler and SDK target interactions work at a low level.
Second, the Extending The SDK <https://llvm-mos.org/wiki/Extending_SDK>
guide, which describes how our SDK's CMake build is structured.

Given the similarities between the C64 and the Mega65, using it as a base
makes sense. There's already a mechanism in the SDK for one target to be
based on another "incomplete target"; this is in an attempt to reduce code
duplication between the targets. For example, all targets derive from the
"common" target, and if you look at the NES targets, you can see an actual
tree of targets based on board and mapper type.

Depending on the specifics of the differences between the Mega65 and the
C64, it may be possible to use this mechanism directly, otherwise, it may
be best to extract out the common bits into a "commodore" target, then make
both C64 and Mega64 derive from it.

Please do at least skim through those guides; they should get you up to
speed on how everything ticks on the inside. Afterwards, feel free to drop
by our Slack Group
<https://join.slack.com/t/llvm-mos/shared_invite/zt-t88fyh4i-EGCLe~MSlHdz3~h~yYHgFA>;
I'm online frequently to answer questions, and there's other folks there
who've gone through the same process.

Take care!
-- Dan

mlund:
Hi Dan, thanks for information - I completely missed the porting guide which looks exactly what I was after. Will get back when I run into trouble. Thanks / Mikael

PR #45, opened 2022-06-23
Create common commodore target

mlund:
In preparation for upcoming commodore targets, _e.g._ mega65 (#44), this splits out common features to a general commodore target. I kept the `c64` target name for backwards compatibility, but perhaps it's better to use `commodore-64` to follow the convention used for the `nes-nrom-*` targets(?). Ping @mysterymath.


mos-platform/c64/CMakeLists.txt
```diff
@@ -1,30 +1,10 @@
-platform(c64 COMPLETE HOSTED PARENT common)
```
mysterymath, line 1: I think we still need the HOSTED tag here, since the target has a real operating system (although right now that's just putchar, getchar, and malloc).

mysterymath (changes requested):

mos-platform/c64/CMakeLists.txt
mlund: ok, fixed that.

PR #46, opened 2022-06-26
Add mega65 target

### LINKED ISSUE #44: MEGA65 target

mlund:
This adds a MEGA65 target and should further be compatible with the unreleased Commodore 65 computer. This is a basic but functional implementation; memory regions could likely be optimised to better use the expanded memory of the mega. Closes #44.

- [x] Add mega65 target
- [x] Update README 
- [x] Resolve issue with `printf` (`examples/hello-new`)

~~Memory is allocated in $2001-$BFFF and `__stack` is set to $BFFF. However, this seem to give problems with `printf` that prints the same number regardless of input. Lowering the stack location solves this issue.~~
Currently the SYS command does not map bank 0's $8000-$C000 region which caused the above issue. Ping @mysterymath


mos-platform/CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ add_subdirectory(common)
 add_subdirectory(atari8)
 add_subdirectory(commodore)
 add_subdirectory(c64)
+add_subdirectory(mega65)
```
mlund, line 41: Sort alphabetically?

mos-platform/mega65/basic-header.S
```diff
@@ -0,0 +1,10 @@
+; MEGA65 BASIC header that jumps to _start.
+
+.section .basic_header,"aR",@progbits
+#	.short next_line	; address of next BASIC line
+        .byte 0x0a, 0x20, 0x0a, 0x00, 0xfe, 0x02, 0x20, 0x30, 0x00 ; 10 BANK 0
+        .byte 0x15, 0x20, 0x14, 0x00, 0x9e, 0x20 ; 20 SYS
+	.mos_addr_asciz _start, 4	; the location of the _start symbol, in 4 decimal
```
mlund, line 7: is there a more elegant way of doing this?

mysterymath: Not that I know of; this seems pretty idiomatic for BASIC headers.

mysterymath (approved):
Looks good to me!

PR #49, opened 2022-07-05
Initial support for the dodo platform

atn34:


mos-platform/dodo/api.h
```diff
@@ -0,0 +1,144 @@
+#pragma once
```
mysterymath, line 1: The `#pragma once` here seems redundant with the include guards. Was there a particular reason to use both?

mos-platform/dodo/crt0.s
```diff
@@ -0,0 +1,10 @@
+.section .init.40,"axR",@progbits
+	; make sure flags are in a sane state
+	cld
+	cli
+	; tell dodo where the software stack pointer is
+	lda #__rc0
+	sta $00
+	; Init hardware stack pointer
+	ldx #$FF
+	txs
```
mysterymath, line 10: nit: No newline at end of file.

mysterymath (approved):
Generally looks good to me, sans a couple nitpicks.

We generally don't require any runtime testing for SDK targets, given the crapshoot that is scripting support for common system emulators. We test common SDK libraries through the simulator target, and just make sure that the suite of examples that ship with the SDK successfully compile.

You may want to consider adding a dodo-specific example to the SDK, but it's not required.

mos-platform/dodo/api.h
atn34: I got this from another project I've worked on where I guess the idea is that the include guard is more portable but pragma once is faster on some compilers (allegedly), so each has a benefit. Looking through some headers in other platforms I see a possible preference for include guards, so I'll go with that. I don't have a strong opinion.

PR #68, opened 2022-08-08
Add Commander X16 target

### LINKED ISSUE #51: target Commander x16 (similar to c64)?
any chance we can get this to target commander x16?
[https://www.commanderx16.com](https://www.commanderx16.com)

mlund:
Rudimentary `cx16` target with some hardware register support as defined in cc65. Thanks to Slack user **serentty** for making the original x16 linker file (slightly modified here). The cx16 uses the same KERNAL jump table as the C64 as well as the same BASIC areas, making the two targets almost identical. Closes #51.

Notes:
- A Commander X16 emulator is [available](https://github.com/commanderx16/x16-emulator).
- [x] Does `-mpu=mos65c02` in `clang.cfg` have any effect? I did not detect a binary size difference on simple examples (and seem to remember that mcpu silently accepts any string). See #70.
- [x] Please check licence inclusion in file headers as the `.h` files were taken from cc65. I tried to follow the same scheme as for the c64 target.
- [ ] Check available zero page memory, see [here](https://www.c64-wiki.com/wiki/Commander_X16). From this information, it seems `0x22-0x7f` is free and almost matches the c64 (a few more bytes could be added). ZP is however defined in the parent `commodore.ld` file. Should this be moved to derived targets, as `c64`, `mega65`, and now `cx16` likely all differ slightly? My opinion is that it's probably OK for now, since the X16 and MEGA65 projects could still be subject to memory map changes.


mos-platform/cx16/_6522.h
```diff
@@ -0,0 +1,73 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
```
mysterymath, line 1: I just finished relicensing the SDK, so we're only accepting contributions made available under the new license. Accordingly, this should read:

```
// Copyright 2022 LLVM-MOS Project
// Licensed under the Apache License, Version 2.0 with LLVM Exceptions.
// See https://github.com/llvm-mos/llvm-mos-sdk/blob/main/LICENSE for license
// information.
```

mos-platform/cx16/cx16.h
```diff
@@ -0,0 +1,325 @@
...
+#define COLOR_PINK              0x0A
+#define COLOR_LIGHTRED          COLOR_PINK
+#define COLOR_GRAY1             0x0B
+#define COLOR_GRAY2             0x0C
+#define COLOR_LIGHTGREEN        0x0D
+#define COLOR_LIGHTBLUE         0x0E
+#define COLOR_GRAY3             0x0F
+
+/* TGI color defines */
+#define TGI_COLOR_BLACK         COLOR_BLACK
```
mysterymath, line 123: I consider TGI bloat (but I'm open to having my mind changed), so I don't really intend to support it. Accordingly, we can remove these until we're actually trying to get TGI working (currently slated for never).

mysterymath:
> * [ ]  Does `-mpu=mos65c02` in `clang.cfg` have any effect? I did not detect a binary size difference on simple examples (and seem to remember that mcpu silently accepts any string).
File a bug if you would for clang `-mcpu`. The `llc` backend compiler accepts `--mcpu=mos65c02`, and it'll fail if you try `--mcpu=foo`. It looks like something isn't getting piped through properly.

mysterymath:
Also, thanks for doing this! I was actually just about to get started on #51, but you beat me to it! Great feeling, the project has legs and is starting to walk!

mos-platform/cx16/_6522.h
mlund: Fixed

mos-platform/cx16/cx16.h
mlund: TGI removed

mlund:
> > * [ ]   Does `-mpu=mos65c02` in `clang.cfg` have any effect? I did not detect a binary size difference on simple examples (and seem to remember that mcpu silently accepts any string).
> 
> File a bug if you would for clang `-mcpu`. The `llc` backend compiler accepts `--mcpu=mos65c02`, and it'll fail if you try `--mcpu=foo`. It looks like something isn't getting piped through properly.

OK, will do. I keep the `-mpu=mos65c02` line for this target as the syntax seems correct albeit has no effect at the moment.

mlund:
> Also, thanks for doing this! I was actually just about to get started on #51, but you beat me to it! Great feeling, the project has legs and is starting to walk!

I've had a lot of fun with llvm-mos, so thank you!

mysterymath (approved):

mysterymath:
By the way, would you like to put together an announcment message on the Commander X16 forums or Discord? It'd be a good idea to let that community know this is available, and it's your work, so feel free to do the honors. I can also take care of this if that sounds unfun.

Serentty:
@mlund Oh hey, that’s me! It’s cool to know something based on my very minor effort has been merged! Thanks for finishing it.

PR #73, opened 2022-08-10
Add CPU definition to mega65 target

mlund:
The 65ce02 is currently the closest match to the derived GS4510 used in the mega65.


mysterymath (approved):

Serentty:
@mlund It seems that the MEGA65 target does not produce binaries which will load on the original Commodore 65. What would need to be done to add a Commodore 65 target?

PR #80, opened 2022-10-04
SDK port for CP/M-65

davidgiven:
Hello,

I recently did a port of CP/M to the 6502; this SDK port produces portable binaries for it. The only interesting thing is that the binaries have to be relocatable, so the binaries are generated as relocatable ELF and then there's an external tool for converting the ELF file to CP/M-65 binaries. This is pretty ropey as I'm not great with ELF and I'd appreciate some feedback (e.g. it doesn't work if the binaries are compiled with debugging, because debug sections are apparently generated as PROGBITS... what's the right way to detect sections which should end up in the output file?). There's also a CP/M system call library provided.

If you're interested it'd be great to get the upstreamed --- please let me know what you need.

Some questions:

- do you have guidelines on coding conventions, license headers etc? (A clang-format file?)
- there's an option for running a tool automatically after ELF generation (`-fpost-link-tool`) but I couldn't figure out how to work it. This seems to require outputting another ELF file, right? Is there a way to use an external tool to do the ELF-to-binary conversion, or is this always assumed to be done by the linker?
- there's an idiom where pulling in an object file contributes a chunk to the init sequence (e.g. `https://github.com/llvm-mos/llvm-mos-sdk/blob/main/mos-platform/common/crt0/copy-zp-data.c`). What prevents the init section from being discarded as there are no references to it? Is this done by magic because of the section name?
- how do I import zero page symbols into an asm file and let the compiler know they're zero page?
- I also added some ctype functions, just because one of the programs I was compiling needed them.

See https://github.com/davidgiven/cpm65.

Thanks!


mysterymath:
Hey, neat project! Thanks for taking the time to put together a PR!

> * do you have guidelines on coding conventions, license headers etc? (A clang-format file?)

Nah, we've been lazy about that for the SDK so far. It's starting to get a bit wooly, I might put something more automated in later and do a clean sweep. If anything is too horrifying I'll yell, otherwise I wouldn't worry too much about it.

> * there's an option for running a tool automatically after ELF generation (`-fpost-link-tool`) but I couldn't figure out how to work it. This seems to require outputting another ELF file, right? Is there a way to use an external tool to do the ELF-to-binary conversion, or is this always assumed to be done by the linker?

When our custom `OUTPUT_FORMAT {}` linker extension is used, the linker will output both an object file in the given format (`the_main_file`) and an auxiliary ELF file (`the main_file.elf`). `-fpost-link-tool` was designed with that in mind, but I should try to get it to work if the main output file is ELF. Would you file a bug against llvm-mos for this?

> * there's an idiom where pulling in an object file contributes a chunk to the init sequence (e.g. `https://github.com/llvm-mos/llvm-mos-sdk/blob/main/mos-platform/common/crt0/copy-zp-data.c`). What prevents the init section from being discarded as there are no references to it? Is this done by magic because of the section name?

These are given the SHF_RETAIN ELF flag in the assembler: `.section name,"...R..."`. This makes them immune to linker GC.

> * how do I import zero page symbols into an asm file and let the compiler know they're zero page?

https://llvm-mos.org/wiki/Assembler_relaxation

> * I also added some ctype functions, just because one of the programs I was compiling needed them.

Not really a question, but thanks! :p

I'll have to take some time to go through it in detail, but first things first: the libelf dependency isn't going to fly, since we compile the SDK for stock Windows. We're overall trying to keep our dependency count small; accordingly, LLVM is the source of most of our complex compiler-ey functionality.

Accordingly, I think the right place to do the binary file generation is inside our LLVM repo. There, you'll have access to the full and excellent ELF libraries provided with LLVM. Although unfortunately, like pretty much everything in LLVM, there's very little documentation. There's already something vageuely similar to extract out Mesen label files from ELF for the NES (the existing post-link tool): https://github.com/llvm-mos/llvm-mos/tree/main/llvm/tools/llvm-mlb. You should be able to port your binary-file generator similarly; then we can ship it as a statically-linked binary that works across platforms.

Another alternative would be to add this functionality to llvm-objcopy; this is a tool explicitly designed for converting from one object file format to another. But, it's pretty complicated, so I wouldn't count too hard on it actually saving any time, especially if you're not super-familiar with the LLVM codebase.

davidgiven:
Thanks!

Re assembler relaxation: the doc's not quite clear; is it the _linker_ which does the relaxation, rewriting the code to use smaller instructions? Or does the assembler do this? The examples given use section statements to define zero page addresses in the same file as they're used, so the assembler knows they're zero page addresses, but what about importing symbols defined elsewhere? (e.g. cc65's `.zpimport` instruction.)

Re `-fpost-link-tool`: I think I can work with that, actually. Let me have a play and get back to you, especially if I need to rework the ELF stuff.

mysterymath:
> Re assembler relaxation: the doc's not quite clear; is it the _linker_ which does the relaxation, rewriting the code to use smaller instructions? Or does the assembler do this? 

It's the assembler; we don't yet have a link-time relaxation feature (although LLVM doesn't preclude one from being written!).

> The examples given use section statements to define zero page addresses in the same file as they're used, so the assembler knows they're zero page addresses, but what about importing symbols defined elsewhere? (e.g. cc65's `.zpimport` instruction.)

Hmm, when you put it that way, I don't think there's a way to do it yet. We've been manually using 8-bit modifiers to force zero-page addressing for cases like that, but that's a decidedly worse solution than `.zpimport`. I don't think anything fundamentally precludes implementing that in our assembler though.

I think a real link-time relaxation feature would supercede this, though; if you could just rewrite the operations based on the real effective addresses, then you'd never need to bother marking symbols as ZP. I'd guess (mostly without evidence) that the two approaches are about the same size; so I think I'd want to try the linker route first. Feel free to file an issue against the compiler project for this.

davidgiven:
Done: see https://github.com/llvm-mos/llvm-mos/issues/228.

BTW, another question: the CP/M APIs all take and return 16-bit parameters in XA. I asked on a bug about how to pass in pointer parameters to inline assembly in XA; but, how would I _return_ a pointer parameter in XA? Currently there are stub functions for converting the calling convention, but I think that doing it all in inline assembly in the header file should produce better code.

mysterymath:
I missed that you wanted to to this in a header via a macro; that's trickier, since you'd need to stay in an expression context. You could probably use inline assembly to do the call and construct the return, but you'd need some wat to make that the expression value, probably involving deep compiler torture like statement expressions.

See https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html

davidgiven:
_High_ byte in X, so it does actually match the integer calling convention --- but the issue is that CP/M wants to take and pass back pointers in XA, which llvm-mos wants to pass in an imaginary register.

davidgiven:
_facepalm_

mysterymath:
> This is pretty ropey as I'm not great with ELF and I'd appreciate some feedback (e.g. it doesn't work if the binaries are compiled with debugging, because debug sections are apparently generated as PROGBITS... what's the right way to detect sections which should end up in the output file?). 

From the ELF standard, PROGBITS actually just means "data for this section is in the binary, but the meaning is entirely program-specific."

For determining whether a section is expected to occupy memory at runtime, the attribute is the section flag SHF_ALLOC. This indicates that the section should be assigned a virtual address, corresponding to it's runtime address. If the section is PROGBITS, then the binary contains data expected to be loaded at that address; if it's NOBITS, then it doesn't. There are also other kinds of SHF_ALLOC sections understood by ELF, like unwind tables etc., but we shouldn't really have any of those to worry about on the 6502.

davidgiven:
_facepalm_ (again). I can't believe I didn't spot that...

I actually think I can simplify the tool drastically by working with program headers rather than sections, which will let me remove the libelf dependency.

utils/elftocpm65/elftocpm65.cc
```diff
@@ -1,23 +1,35 @@
 #include <fcntl.h>
-#include <unistd.h>
 #include <fmt/format.h>
-#include <libelf.h>
-#include <string>
 #include <fstream>
-#include <vector>
 #include <set>
+#include <string>
+#include <unistd.h>
```
mysterymath, line 6: Neither `unistd.h` nor `elf.h` will be present on a MSVC install on Windows, so this utility is still POSIX-specific.

But, in retrospect, I don't want to burst your bubble, so I think that's OK, especially since it doesn't add anything to our install instructions on most POSIXen. Just add some CMake to prevent the build and install of this utility on Windows, and we can call it a day.

mos-platform/cpm65/cpm.h
```diff
@@ -95,36 +95,74 @@ extern uint8_t cpm_errno;
...
-/* 16 */ extern uint8_t cpm_close_file(FCB* fcb);
-/* 17 */ extern uint8_t cpm_findfirst(FCB* fcb);
-/* 18 */ extern uint8_t cpm_findnext(FCB* fcb);
-/* 19 */ extern uint8_t cpm_delete_file(FCB* fcb);
-/* 20 */ extern uint8_t cpm_read_sequential(FCB* fcb);
-/* 21 */ extern uint8_t cpm_write_sequential(FCB* fcb);
-/* 22 */ extern uint8_t cpm_make_file(FCB* fcb);
-/* 23 */ extern uint8_t cpm_rename_file(RCB* rcb);
+/* 15 */ extern uint8_t cpm_open_file_i(uint16_t fcb);
+         static uint8_t cpm_open_file(FCB* fcb)
```
mysterymath, line 107: Nice, these should just get inlined away!

There's a sliiight smell here, in that cpm_open_file in one C file won't == cpm_open_file in another. Since we use LTO, you could just move these wrappers to a C file, rather than a header. The compiler should still inline them all the same, since it has a global view of all calls and definitions.

mos-platform/cpm65/cpm.h
```diff
@@ -95,36 +95,74 @@ extern uint8_t cpm_errno;
 /*  6 */ extern uint8_t cpm_conio(uint8_t b);
```
mysterymath, line 95: Most of these routines should probably be marked with `__attribute__((leaf))`. This tells the compiler that the underlying assembly routine can't call back into the main program visible at link time via LTO. Otherwise, the compiler would be forced to assume that your assembly routines could do anything, up to including calling `main`. That means it would have to consider any functions on the path from `main` down to these call recursive, and use a cc65-style stack for them.

Marking most/all of these as leaves should allow your C utilities to benefit from llvm-mos's static stack and global zero page allocation capabilities. You only would need to do this for the `_i` variants; anything the compiler can't see through.

mos-platform/cpm65/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
+platform(cpm65 HOSTED PARENT common)
+
+if(NOT CMAKE_CROSSCOMPILING)
+  return()
+endif()
+
+install(FILES
+  cpm.h
+#  cbm.h
```
mysterymath, line 9: Nit: no commented code.

mos-platform/cpm65/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+  cpm.S
+  bios.S
+  comhdr.S
+  pblock.S
+  putchar.c
+  stack.S
+  registers.S
+)
+merge_libraries(cpm65-c
+	common-c
```
mysterymath, line 31: nit: indentation

mos-platform/cpm65/clang.cfg
```diff
@@ -0,0 +1,3 @@
+-D__CPM65__
```
mysterymath, line 1: You might also want to set `-mlto-zp` to let the compiler know how many zero page locations it's allowed to play with, not including the imaginary register bank. This allows it to lift global variables, callee-saved-registers, and local variables to the zero page, if it's profitable to do so.

mos-platform/cpm65/cpm.h
```diff
@@ -0,0 +1,183 @@
+#ifndef CPM_H
+#define CPM_H
+
```
mysterymath, line 3: Add the C++ "extern C" headers so this header can work with C++ too. See various other headers in the SDK.

mos-platform/cpm65/registers.S
```diff
@@ -0,0 +1,24 @@
+.macro register n, m
```
mysterymath, line 1: Might want to add a comment on why this is used rather than the usual linker script mechanism. I assume it's something to do with the relocations?

mos-platform/cpm65/stack.S
```diff
@@ -0,0 +1,7 @@
+.global __stack
```
mysterymath, line 1: Is there any way to have the stack start at the end of memory and grow downward?

Note that heap created by malloc assumes the usual C layout: text/data, then heap growing up, followed by free space, followed by stack growing down. This will cap both the stack and heap size to 2K.

utils/elftocpm65/elftocpm65.cc
```diff
@@ -0,0 +1,285 @@
+#include <fcntl.h>
+#include <fmt/format.h>
```
mysterymath, line 2: Probably shouldn't have a fmt dep either; at the very least, this will break our builders.

davidgiven: Sob. Removed. I've also changed it to work in post-link-tool mode, which makes argument parsing easier, so I got rid of getopt too.

mos-platform/cpm65/registers.S
davidgiven: Comment added (it's so the linker can discard unused registers, which seems to happen a lot). (Especially in hand-written machine code.)

mos-platform/cpm65/stack.S
davidgiven: Technically yes, but I'd rather not? The issue here is that the command processor lives at the top of memory, allowing it to be overwritten by programs which need the extra 2kB of RAM (and, for CP/M-65, the zero page it consumes). Putting the stack high commits us to either always overwriting the command processor or else never overwriting it. I'll also note that the heap is actually above the stack in my model.

What I was planning (not done yet) was to use linker magic so that if a program refers to the heap, then the command processor gets overwritten and a warm-boot command gets inserted into the shutdown code --- this tells the BDOS to reload the command processor. Otherwise, the program just returns normally.

mos-platform/cpm65/cpm.h
davidgiven: Done.

mos-platform/cpm65/clang.cfg
davidgiven: I did try this, but it ended up producing bad code and my programs didn't work. I'm still investigating why.

mos-platform/cpm65/CMakeLists.txt
davidgiven: Done.

mos-platform/cpm65/CMakeLists.txt
davidgiven: Done.

mos-platform/cpm65/cpm.h
davidgiven: Done.

mos-platform/cpm65/cpm.h
davidgiven: Done.

utils/elftocpm65/elftocpm65.cc
davidgiven: Would it be feasible to just include a copy of elf.h? There's an Apple version which is CDDL, which (IIRC) should be compatible with Apache.

davidgiven:
I've hopefully addressed most of the comments, but there's still some work to do...

Some questions:

- is there a way to have an alternative `main()` function? I'd like to have my CRT startup sequence call, say, `cpmmain()`, which can either be implemented by the user or which is implemented in the runtime to parse command line arguments and call the conventional `main(argc, argv)`. To do this I'd need to replace the `common-crt0-o` file defined in `mos-platform/common/crt0` but I can't tell how that file gets pulled in --- it all happens through cmake magic.
- is zp-data currently used by anything? I think the problem I'm having with LTO is due to this not being set up correctly, but I can't actually tell what's going on --- as LTO happens inside the linker I have limited visibility of what sections it's actually creating...
- how do I set the top-of-heap address? I see `__set_heap_limit`, but I have a bit of a suspicion that calling this in my startup code will pull in the entire memory allocator, which would be bad if it's not being used.

utils/elftocpm65/elftocpm65.cc
mysterymath: I did a quick scan through the license; that should be fine. I think we're actually using MinGW, so there shouldn't be any compatibility problems. If there are, our builders will break, I'll roll it back, and we can fix things from there.

mos-platform/cpm65/stack.S
mysterymath: Ah, makes sense.

mysterymath:
> * is there a way to have an alternative `main()` function? I'd like to have my CRT startup sequence call, say, `cpmmain()`, which can either be implemented by the user or which is implemented in the runtime to parse command line arguments and call the conventional `main(argc, argv)`. To do this I'd need to replace the `common-crt0-o` file defined in `mos-platform/common/crt0` but I can't tell how that file gets pulled in --- it all happens through cmake magic.

The rule we've been maintaining in the SDK is that dependent targets can only add things to the targets they depend on, not replace them. This keeps things simple, but requires that the base targets only include truly generic code; everything else needs to be either an optional library that can be merged in or a stub that can be filled.

That being said, up until right now, JSR-ing to main was generic, since none of our targets had the ability to receive arguments through main. Now that one needs this, we'd make the `.call_main` section in `common/crt0/crt0.S` something weak.

I think it would suffice to move it into the `crt` collection of libraries, have it define a weak symbol like `__call_main`, then use an undefined reference to it in `crt0.S`. That should only pull in the `JSR main`  if the user doesn't define a strong version of that symbol. This is roughly how `__after_main` currently works.

> * is zp-data currently used by anything? I think the problem I'm having with LTO is due to this not being set up correctly, but I can't actually tell what's going on --- as LTO happens inside the linker I have limited visibility of what sections it's actually creating...
You can try passing `-Wl,--lto-emit-asm` to the linker; this will cause it to spit out the full assembly it generates. You can also take a look at `-Wl,-Map=map.file`, the two together give a good purchase on what the linker is doing RE lto. Also `-Wl,-save-temps`, if you need all the gratuitious detail.

As for ZP data, it's used whenever anything in the `.data` section is lifted to the zero page. These variables have initializers, so a copy routine is used to set them at program start, just like the regular data section. This incurs an additional cost, but the code generator incorporates that into its cost model, and it tries to only lift variables where the access frequency is worth the additional loop to set them up.

It's also a convenience for library authors, since it allows registering ZP locations to get set with initial values without needing to schedule a routine to run at initialization time. A single `__copy_zp_data` call can initialize all ZP used by the program in a single loop. Note that currently, the implementation of that function is hot garbage (just a memset). There's some open compiler work to make it better, but it's rather difficult, and it may be a good idea to rewrite it in assembly in the meantime.

> * how do I set the top-of-heap address? I see `__set_heap_limit`, but I have a bit of a suspicion that calling this in my startup code will pull in the entire memory allocator, which would be bad if it's not being used.

There's no fixed way to do this at present. The heap size defaults to 4096, and the limit only produces allocation failures; the heap doesn't actually grow to that size until that much memory is allocated.

mos-platform/cpm65/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+  cpm.S
+  cpm-wrappers.c
+  bios.S
+  comhdr.S
+  pblock.S
+  putchar.c
+  stack.S
+  registers.S
+)
+merge_libraries(cpm65-c
```
mysterymath, line 29: This should be automatically done by the add_platform_library directive. See if it still works with the `merge_libraries` removed.

mos-platform/cpm65/link.ld
```diff
@@ -0,0 +1,70 @@
...
+	} :relo
+
+	.symtab : {
+	} :symtab
+
+	PROVIDE (cpm_ram = __heap_start);
+
+}
+
+/* The result of this is actually irrelevant --- the elftocpm65 tool will
```
mysterymath, line 62: Thanks, I hate it. :) Although there are worse hacks in llvm-mos.

No action required, but please do file an issue at least to make -fpost-link-tool work properly in this case.

mysterymath (approved):

mysterymath:
Deepest apologies for merging this too early; I shouldn't have marked this as approved, since it wasn't ready, and I took that "Approved" as a sign that I had just forgotton to merge this. This still needs at least the ELF header to be added; it looks like that caused failures on OSX as well (?).

davidgiven:
No, that's cool. I've done the same... I need to sort out why .zp.data isn't working, anyway (now possible thanks to your fix for -g!).

davidgiven:
I've done some playing with LTO. The reason why things aren't working is because my .zp.data and .zp.rodata sections aren't being emitted, and I have no idea way. Is there any chance you could take a quick look at my linker file and sanity test the .zp sections, please, in case there's something stupid in there I'm looking past? Link to the relevant bit is here: https://github.com/davidgiven/llvm-mos-sdk/blob/main/mos-platform/cpm65/link.ld#L36

(The `:init` is to cause that data to be emitted to a new program header, because lld gets very upset if you try to rewind the virtual address inside a program header; the BYTE(0) is to ensure that the section size is non-zero and will be emitted.) (Which it is --- one byte long...) (And --emit-asm is showing that there really should be data there.)

Thanks!

mysterymath:
I'd try compiling it with `-Wl,-save-temps` and see if the `<whatever>.lto.o` object file actually has those sections in it. That's the object file actually produced by the LTO compile, and it's included in the link just as if it had been specified on the command line.

There's roughly two classes of problems I can think of: that it doesn't make it out of LTO, and that it doesn't make it out of the linker. Your linker script looks OK to me; so I don't really have any off the cuff explanations as to why this is happening. But narrowing down exactly when and where it's happening should help.

mysterymath:
Noooope I've got it.

The rule for linker scripts is "first matching wildcard wins".
The noinit ZP sections include file contains: `*(.zp .zp.*)`
Since that appears in your linker script before the rodata one, it'll gobble up everything in the rodata section.

That's unfortunate from a linker script ergonomics POV, but noinit is the usual semantics for ZP variables/sections, so I think it's probably still the right call overall. But that does mean that it has to be the last ZP section in the linker script.

davidgiven:
Oh, good grief.

Thanks. That should (I hope) be the last major blocker.

PR #83, opened 2022-10-11
CP/M-65 port (take two)

davidgiven:
This should be ready to go now.

There are a few things missing, like argument parsing and the heap, but if you're willing to do without you can write perfectly useful CP/M-65 programs (a couple of my standard utilities are in C). So if you think this is worth having I'd like to merge this now and deal with the rest later.

I've added basic copyright notices to all my files to make the license and author clear --- let me know if these are okay or if you want them changed.


mos-platform/cpm65/bios.S
```diff
@@ -0,0 +1,20 @@
+; Sets up the BIOS entrypoint.
+;
+; © 2022 David Given
+; This file is part of the llvm-mos-sdk project and is redistributable under
+; the terms of the Apache 2.0 license. See the LICENSE file in the project root
```
mysterymath, line 5: Can these files instead be licensed using the "Apache 2.0 license with LLVM exceptions"? This is what most of the rest of our SDK uses, as it allows the copies of the object files included with statically-linked user binaries to be redistributed without including a copy of the Apache license.

davidgiven: Yes, totally. Done.

davidgiven:
Woo, thanks!

PR #87, opened 2022-10-21
Add code for cbm_k functions in cbm.h

cnelson20:
- Added assembly source for kernal wrapper functions
- Removed an ifdef that always evaluated to false to add back function headers in cbm.h
- Removed `__fastcall__` tag from cbm_k functions
- Added .s files for the kernal wrappers to the commodore libc


mos-platform/commodore/cbm_k_acptr.s
```diff
@@ -0,0 +1,10 @@
+.text
+
+#include <cbm_kernal.inc>
+
+;
+; unsigned char cbm_k_acptr (void);
+;
+.global cbm_k_acptr
+cbm_k_acptr:
+	jmp ACPTR
```
mysterymath, line 10: nit: Add newlines at the end of files.

mos-platform/commodore/cbm_k_acptr.s
```diff
@@ -0,0 +1,10 @@
+.text
+
+#include <cbm_kernal.inc>
```
mysterymath, line 3: This is actually a comment, so cbm_kernal.inc isn't included.

The reason this still works is that the assembler treats ACPTR as an undefined external symbol, and the linker is able to fix it up later to point to the actual ACPTR. This will break if the user defines a symbol named ACPTR, which the C standard requires libc to accept.

You can fix this by using the assembly language include directive: `.include "cbm_kernal.inc"`. Alternatively, you can let the linker handle this, but use the reserved symbol `__ACPTR`, which users cannot legally define.

The same logic applies to most of the files in this PR.

mos-platform/commodore/cbm_k_acptr.s
```diff
@@ -0,0 +1,10 @@
+.text
+
+#include <cbm_kernal.inc>
+
+;
+; unsigned char cbm_k_acptr (void);
+;
+.global cbm_k_acptr
+cbm_k_acptr:
```
mysterymath, line 9: Optional: if you have time, it would be better to write as many of these functions as possible in C. This would allow thin wrappers like this to be completely inlined at the point of use: a call to cbm_k_acptr would just become jsr ACPTR inside the user code.

In C, it would look like this:
```C
unsigned char __ACPTR(void);
unsigned char cbm_k_acptr(void) {
  return __ACPTR();
}
```

No action required; I can always take care of this later, as time permits.

cnelson20:
- Converted some of the source files to C so that they can be inlined
- updated labels to use __LABEL 's

mysterymath (approved):
Alright, LGTM. Thanks for putting this together! This makes the commodore targets much more useful out of the box.

PR #89, opened 2022-11-01
Check for the batch file version for mos-clang platforms

jroweboy:
On windows, cmake doesn't adhere to PATHEXT for determining what is executable, and has its own hardcoded list. So while `find_program` will happily append `.exe` when looking for the file, it will not automatically add `.bat`.

Since the `mos-clang` wrappers for each configuration are in batch files, this works around the limitation by searching for the specific compiler with `.bat` appended to it, allowing `find_program` to work as expected.


mysterymath:
Ah, thanks for finding and fixing this. LGTM.

PR #92, opened 2022-12-06
add atari8-stdcart target

cwedgwood:


mos-platform/atari8-stdcart/stdcart.ld
```diff
@@ -0,0 +1,42 @@
+/* Provide imaginary (zero page) registers. */
+__rc0 = 0x80;
+INCLUDE imag-regs.ld
+ASSERT(__rc31 == 0x9f, "Inconsistent zero page map.")
+
+ASSERT(__cart_rom_size == 1 << LOG2CEIL(__cart_rom_size),
+       "CART-ROM size must be a power of 2.")
+
+MEMORY {
+  zp : ORIGIN = __rc31 + 1, LENGTH = 0x100 - (__rc31 + 1)
```
mysterymath, line 10: Do cartridges really not depend on any of the OS zero page regions? Does any action need to be taken on initialization to disable OS interrupt handlers that might affect this region?

mos-platform/atari8-stdcart/syms.s
```diff
@@ -0,0 +1,4 @@
+; rom size, only 8 and 16 are support right now
```
mysterymath, line 1: nit: "in kibibytes"

mos-platform/atari8-stdcart/stdcart.ld
```diff
@@ -0,0 +1,42 @@
...
+                   __zp_data_size : {} >c_rom_vma
+  .zp.bss (NOLOAD) : {
+    INCLUDE zp-bss-sections.ld
+  } >zp
+  INCLUDE zp-bss-symbols.ld
+  INCLUDE zp-noinit.ld
+  .bss (NOLOAD) : { INCLUDE bss-sections.ld } >c_ram_vma AT>c_ram_lma
+  INCLUDE bss-symbols.ld
+  .noinit (NOLOAD) : { INCLUDE noinit-sections.ld } >c_ram_vma AT>c_ram_lma
+
```
mysterymath, line 41: nit: remove extra line

mos-platform/atari8-stdcart/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+
+install(FILES
+  cart-rom-fixed.ld
+  link.ld
+  stdcart.ld
+TYPE LIB)
+
+# do not create crt0, that will precedence over the common version
+# which we depending on; rather create something specific
+add_platform_library(atari8-stdcart-c
```
mysterymath, line 15: This does need to be in crt0, since it's necessary to enter even a freestanding C environment. The contents of the crt0.o defined in the parent atari8 target should automatically be merged into a platform library of the same name defined here in the child.

mos-platform/atari8-stdcart/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+  syms.s)
+merge_libraries(atari8-stdcart-c
```
mysterymath, line 17: Similarly, it's not a good idea to run around the atari8 target like this; the policy has been that derived targets add to, but to not alter, their parents.

Accordingly, it'll be a bit of work, but we should really break out the common bits of the `atari8` target into an `atari8-common` target. Then atari8 and atari8-stdcart would be thin descendants that fill out the specifics of that configuration. This is the same relationship that `c64` and `cx16` have to commodore, and the various `nes` targets have to `nes`.

mysterymath (changes requested):

mos-platform/atari8-stdcart/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+if(NOT CMAKE_CROSSCOMPILING)
+  return()
+endif()
+
+install(FILES
+  cart-rom-fixed.ld
+  link.ld
+  stdcart.ld
+TYPE LIB)
+
```
mysterymath, line 12: Since this is a ROM target, you'll need some code to copy the `.data` and `.zp.data` segment initializers from ROM to RAM when the program starts. This is provided via a libcrt0.a library:

```
add_platform_library(atari8-stdcart-crt0)
merge_libraries(atari8-stdcart-crt0 common-copy-data)
```

mysterymath (changes requested):

mos-platform/atari8-stdcart/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+# TODO; refactor out atari8 targets
+add_platform_library(atari8-stdcart-crt0 syms.s)
+merge_libraries(atari8-stdcart-crt0
+  atari8-crt0
+  common-copy-data
```
mysterymath, line 17: This causes two copies of the __copy_zp_data function to be present in this library: one inside common-copy-data (it does both), and one inside the merged atari8-crt0.

There's two ways to fix this:
1) Do the refactor to break out an atari8-common; then atari8 can just copy zp data, and atari8-stdcart can copy both zp data and regular data.
2) Make common-copy-data only copy .data, and not include ZP automatically. This is a breaking change, so you'd need to edit all other uses of common-copy-data and add common-copy-zp-data too. If you go this route, you should also update the porting wiki to reflect the new semantics of common-copy-data.

mysterymath (approved):

mos-platform/atari8-stdcart/copy-data.c
```diff
@@ -0,0 +1 @@
+../common/crt0/copy-data.c
```
mysterymath, line 1: We can't check symlinks into the SDK, or it won't build on Windows. Instead, just use this path in the CMakeLists.txt to atari8-stdcart; the oddness of that should be a signal that this needs cleaned up too.

cwedgwood: good point

PR #94, opened 2022-12-17
[C64] Increase available RAM to 50KB by swapping out BASIC ROM

OskarLinde:
On the C64, there is very little reason to keep the BASIC ROM mapped when executing machine code. Switching out the BASIC ROM frees up $A000-$BFFF, and opens up contiguous access to the $C000-$CFFFF area for a total of 12KB of additional RAM.

In the future it would be nice to also have a configuration option to switch out the KERNAL ROM ($E000-$FFFF) and reimplement the KERNAL routines in software so you only use what you need.

### DIFF
--- a/mos-platform/c64/link.ld
+++ b/mos-platform/c64/link.ld
@@ -6,13 +6,16 @@
  */
 
 MEMORY {
-    ram (rw) : ORIGIN = 0x0801, LENGTH = 0x97ff
+    ram (rw) : ORIGIN = 0x0801, LENGTH = 0xC7FF
 }
 
 INCLUDE commodore.ld
 
-/* Set initial soft stack address to end of BASIC area. (It grows down.) */
-__stack = 0x9FFF;
+INPUT(unmap-basic.o)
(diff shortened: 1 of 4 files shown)


mos-platform/c64/link.ld
```diff
@@ -6,13 +6,15 @@
...
 MEMORY {
-    ram (rw) : ORIGIN = 0x0801, LENGTH = 0x97ff
+    ram (rw) : ORIGIN = 0x0801, LENGTH = 0xC7FF
 }
 
 INCLUDE commodore.ld
 
+INPUT(unmap-basic.o)
+
 /* Set initial soft stack address to end of BASIC area. (It grows down.) */
```
mysterymath, line 16: nit: Update comment.

mysterymath (approved):

mos-platform/c64/link.ld
OskarLinde: Thanks, had missed that. Updated.

PR #96, opened 2022-12-18
Add <limits> for integral types.

pfusik:
See #57.

This is incomplete yet, but should work for all typical uses.
The standard defines `std::numeric_limits` for all default-constructible types. This makes no sense for me, but I'll make it standard-compliant later.
I understand that `float` and `double` are out of scope.

### DIFF
--- a/mos-platform/common/include/limits
+++ b/mos-platform/common/include/limits
@@ -1,7 +1,5 @@
 #pragma once
 
-#include <__type-traits-impl.h>
-
 namespace std {
 
 enum float_round_style {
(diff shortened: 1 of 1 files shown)


pfusik:
Also, I need to look into the `const`/`volatile` thing.

mos-platform/common/include/limits
```diff
@@ -0,0 +1,58 @@
+#pragma once
```
mysterymath, line 1: Use explicit include guards rather than pragma once for the SDK.

mysterymath (approved):

mysterymath (approved):

Issue #99, opened 2023-01-16
Unexpected behavior of CBM KERNAL functions


mlund:
For the snippet below, I get different behavior with cc65 and clang using the VICE emulator. I suspect it might be the calling conventions for the `mos-platform/commodore/*.s` functions (`setnam`, `setlfs`, `chkin`). Merely calling `chkin` causes the screen to scroll on clang. The very useful [cbm functions](https://www.pagetable.com/c64ref/kernal/) were introduced in #86 and #87. Ping @cnelson20.

~~~ C
// cl65 cbm_open.c -o open-cc65.prg --target c64
// mos-c64-clang cbm_open.c -o open-clang.prg
#include <cbm.h>
#include <stdio.h>

const char* filename = "SOMEFILE";

int main() {
    unsigned char s1, s2;
    unsigned char lfn = 8;
    unsigned char device = 8;
    unsigned char secondary_addr = 0;

    cbm_k_setlfs(lfn, device, secondary_addr);
    cbm_k_setnam(filename);

    s1 = cbm_k_open(); // Expected = 0; Clang = 199
    //s2 = cbm_k_chkin(lfn); // Expected = 0; on clang this causes the screen to scroll
    printf("open status = %d\n", s1);
    return 0;
}
~~~

mlund:
For [`setlfs`](https://www.pagetable.com/c64ref/kernal/#SETLFS):
- setlfs reference:
  - A = lfn
  - X = device
  - Y = command (secondary address)
- Wiki: _The bytes composing numeric arguments are passed individually in A, then X, then RC2 to RC15._
- So `.s` looks good to me (`ldy __rc2`).

For [`setnam`](https://www.pagetable.com/c64ref/kernal/#SETNAM):
- setnam reference:
  - A = length of filename
  - X = low order address
  - Y = high order address
- Wiki: _Pointers are passed through RS1 through RS7_
- `cbm_k_setnam.s` seem to use rc2 and rc3(?)

cnelson20:
Thanks for the ping, I'll try to take a look at this.

cnelson20:
I haven't found the root cause of the issue since it seems all the registers are set correctly but when using OPEN on a file the secondary address must be between 2 and 14.

cnelson20:
There was a bug in the implementation of cbm_k_open() and some in your code.

[...]

PR #103, opened 2023-02-17
Fix bug in cbm_k_open function

cnelson20:
When the Commodore OPEN kernal routine returns with the carry flag clear, the OPEN should be treated as a success and the value of the accumulator disregarded. Previously, this was not implemented correctly. Now, if carry flag is clear on return from OPEN, 0 is returned to the caller of `cbm_k_open()` indicating a success. 
Thanks to @mlund for finding this (#99)

### DIFF
--- a/mos-platform/commodore/cbm_k_ckout.s
+++ b/mos-platform/commodore/cbm_k_ckout.s
@@ -1,0 +1,15 @@
+.include "imag.inc"
+.text
+
+;
+; unsigned char cbm_k_ckout(unsigned char FN);
+;
+.global cbm_k_ckout
+cbm_k_ckout:
+	tax
+	jsr __CKOUT
+	ldx #0
+	bcs someerror
+	lda #0
+someerror:
+	rts
--- a/mos-platform/commodore/cbm_k_load.s
+++ b/mos-platform/commodore/cbm_k_load.s
@@ -9,7 +9,7 @@
 cbm_k_load:
 	ldy __rc2
 	jsr __LOAD
-	bcs noerror
+	bcc noerror
 	; error code in a ;
 	ldx #0
 	rts
--- a/mos-platform/commodore/cbm_k_load.s
+++ b/mos-platform/commodore/cbm_k_load.s
@@ -8,4 +8,8 @@
 .global cbm_k_load
 cbm_k_load:
 	ldy __rc2
-	jmp __LOAD
+	jsr __LOAD
+	bcs someerror
+	lda #0
+someerror:
+	rts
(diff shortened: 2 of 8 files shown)


mysterymath (commented):
Would you mind doing the same for:
- `cbm_k_ckout`
- `cbm_k_save`

cnelson20:
I updated `cbm_k_ckout`, `cbm_k_save` in the same way, and `cbm_k_load` and `cbm_k_chkin` for good measure.

mos-platform/commodore/cbm_k_ckout.s
```diff
@@ -0,0 +1,15 @@
...
+.text
+
+;
+; unsigned char cbm_k_ckout(unsigned char FN);
+;
+.global cbm_k_ckout
+cbm_k_ckout:
+	tax
+	jsr __CKOUT
+	ldx #0
```
mysterymath, line 11: The return type is unsigned char, so the value of X at return need not be defined. I think you can pull out the `ldx #0.`

mysterymath (approved):

mos-platform/commodore/cbm_k_load.s
```diff
@@ -8,4 +8,8 @@
 .global cbm_k_load
 cbm_k_load:
 	ldy __rc2
-	jmp __LOAD
+	jsr __LOAD
+	bcs someerror
+	lda #0
```
mysterymath, line 13: It looks like a tax is needed here; since the return is 16 bits.

mysterymath:
No action required, but do note that we have a inline assembly support for the carry register as an output from an ASM snippet. This would allow doing this checking in C.

The advantage would be that the compiler would be able to inline this logic at the point of use; in particular, in case nothing examines the return value, the branching logic would be elided. Not sure how common it is to check or not check the return values though; my experience writing real-world code against the KERNAL is very limited.

cnelson20:
I added the `tax` to cbm_k_load.
Thanks for letting me know about the carry register support. I'll make a second PR to try to use that to allow for inlining.

In my albeit somewhat limited experience programming for the 6502, I've never checked the return value of LOAD, since I use for for data files that I know will be there. With OPEN however, checking the return value to check for errors is necessary since it's used in a more dynamic way.

And I don't think I've ever used SAVE. I always just open a file for writing using OPEN and then writing to it with CHKOUT & CHROUT.

cnelson20:
My implementation of cbm_k_load had a couple of other problems. When there is no error, I needed to shift the contents of XY into AX. When there is an error, load should return the error code.
There isn't a good way to let the user check for the error itself then, other than by checking the carry. It's not often needed, since (in my experience) `LOAD` is used more with static data files that are always expected to be on disk.

mysterymath:
Sgtm; thanks for the PR!

Issue #108, opened 2023-03-29
PET target


williamw4096:
I am interested in adding support for the Commodore PET as a target. I have managed to get all of the example programs to compile and run for a 32k PET, mostly by copying over code from the C64 config and adjusting for the PET's memory map and kernel.

Doing this has already taken me beyond the "Porting" and "Extending SDK" parts of the wiki and hence I am not sure how to test that everything is correctly configured. Here is what I have added so far:
- link.ld defining usable RAM space and stack location. Based on the Commodore link file.
- kernal.S exporting some of the definitions from cbm_kernal.inc
- basic-header.S defines the basic SYS command which runs the program; this is identical to the one for the C64
- pet.inc definitions taken from cc65, akin to c64.inc
- added relevant cmake stuff as per the "Extending SDK" wiki page

PR #109, opened 2023-03-30
Initial support for a PET target

williamw4096:
This adds initial support for the Commodore PET as mentioned in issue #108. Example programs compile and run successfully on a 32k PET, but other amounts of memory may not work. I do not know if there is a way to write the link.ld file in order to support PETs with different amounts of memory or whether multiple targets would be required.


mos-platform/pet/link.ld
```diff
@@ -0,0 +1,19 @@
+/* Commodore PET PRG file linker script.                                                                 
+ *                                                                                                       
+ * Produces a PRG file with a SYS command to start the program.                                          
+ */
+
+MEMORY {
+    ram (rw) : ORIGIN = 0x0401, LENGTH = 0x7BFF
```
mysterymath, line 7: You can use expressions involving symbols to set the length of the RAM; that allows users to customize the layout just by defining a symbol, either in a `.s` file or inline assembly.

The NES targets currently do this; the convention they've used is to take the size of various regions in kilobytes, then do the necessary math based on this. Here, that would be the calculations to set the `LENGTH` of `ram` and the value of `__stack`. You'd also probably want an [`ASSERT` statement](https://sourceware.org/binutils/docs/ld/Miscellaneous-Commands.html) to ensure that the value of the symbol is in range. Finally, you can [`PROVIDE`](https://sourceware.org/binutils/docs/ld/PROVIDE.html) the symbol a default value in case the user doesn't supply one.

mos-platform/pet/link.ld
```diff
@@ -0,0 +1,19 @@
...
+}
+
+INCLUDE commodore.ld
```
mysterymath, line 10: The `commodore.ld` linker script makes the zero page from 0x02 to 0x90 available for use by the compiler. Are those regions free on the PET? (That is, the KERNAL interrupts won't clobber them, and the KERNAL is usable regardless of what is written there?)

If more/less ZP is free, it'd be a good idea to pull the zp stuff out of `commodore.ld` down into the dependent targets:
  - `cx16`
  - `pet`
  - `mega65`
  - `c64`

The `clang.cfg` in the commodore directory also affects this; it tells the compiler how much ZP the linker will later be able to satisfy.

mysterymath (commented):
Just a couple comments, but overall, looks pretty good! Thanks for taking the time to put this together.

mos-platform/pet/link.ld
williamw4096: Unfortunately the last 3 bytes of this space don't look to be usable on the PET. 0x8D-8F are used for the jiffy clock so the KERNAL will overwrite them. There doesn't seem to be much extra free ZP apart from the odd unused byte here and there. Let me know if this means ZP stuff should be pulled down into the targets (and whether that should be implemented as part of this pull request)

mysterymath: Yep, you'll need to do this as part of this PR; otherwise, the PET target could produce random behavior for large programs that use the whole ZP space. That's surprisingly common, since the compiler greedily allocates everything it can to ZP.

mos-platform/pet/link.ld
williamw4096: This should be implemented now, I ran some tests on VICE on lower memory models which fail spectacularly with this version but now work with the inline assembly defining the right memory size.

mysterymath: Looks good!

mlund:
Nice! Perhaps also update `README.md` with the new target.

mos-platform/pet/link.ld
mysterymath: Alright, looks good!

mysterymath:
LGTM!

PR #114, opened 2023-05-18
Update CPM65; add some missing ctype and string functions

davidgiven:
This:

- fixes a bug in the CPM65 headers where one function would corrupt the parameters when called
- updated with some new CPM65 functionality
- added a `cpm_parse_filename` helper function (in pure C)
- adds `isalnum` and `isspace` to ctype
- adds `strrev`, `strspn`, `strpbrk` and `strtok` to strings

### DIFF
--- a/mos-platform/common/c/string.c
+++ b/mos-platform/common/c/string.c
@@ -42,9 +42,10 @@ __attribute__((weak)) int strncmp(const char *s1, const char *s2, size_t n) {
   }
 }
 
-__attribute__((weak)) char *strncpy(char *restrict s1, const char *restrict s2, size_t n) {
+__attribute__((weak)) char *strncpy(char *restrict s1, const char *restrict s2,
+                                    size_t n) {
   char *ret = s1;
-  for (;n ; ++s1, ++s2, --n) {
+  for (; n; ++s1, ++s2, --n) {
     *s1 = *s2;
     if (!*s2) {
       for (++s1, --n; n; ++s1, --n)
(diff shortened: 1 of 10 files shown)


mos-platform/common/c/string.c
```diff
@@ -63,3 +63,70 @@ __attribute__((weak)) char *strrchr(const char *s, int c) {
...
+{
+    size_t len = strlen((const char*)str);
+    size_t i, j;
+    for (i = 0, j = len - 1; i < j; i++, j--)
+    {
+        char a = str[i];
+        str[i] = str[j];
+        str[j] = a;
+    }
+	return str;
```
mysterymath, line 77: nit: decrease indent

mos-platform/common/c/string.c
```diff
@@ -63,3 +63,70 @@ __attribute__((weak)) char *strrchr(const char *s, int c) {
       last = s;
   return (char *)last;
 }
+
+__attribute__((weak)) char* strrev(char* str)
+{
+    size_t len = strlen((const char*)str);
+    size_t i, j;
+    for (i = 0, j = len - 1; i < j; i++, j--)
```
mysterymath, line 71: nit: prefer declaring variables inside loops if only used inside, here and throughout. The llvm-mos stdlib generally targets c99, and even C++ is allowed if ABI-compatible.

mos-platform/common/c/string.c
```diff
@@ -63,3 +63,70 @@ __attribute__((weak)) char *strrchr(const char *s, int c) {
...
+
+    if (string == NULL)
+    {
+        string = savestring;
+        if (string == NULL)
+            return (char*)NULL;
+    }
+
+    char* s1 = string + strspn(string, separators);
+    if (*s1 == '\0')
```
mysterymath, line 120: Use !*s1, for consistency with general use of *s1 as a conditional

mysterymath (approved):
Hey, thanks for putting this together! Always nice to see more std library functions :)

mos-platform/common/c/string.c
davidgiven: I actually had to look up how to declare two variables at once in a for loop. I think I died inside a little. Done.

mos-platform/common/c/string.c
davidgiven: Forgot to run through clang-format. Done.

mos-platform/common/c/string.c
davidgiven: Done.

PR #117, opened 2023-06-16
Provide 45KB RAM for MEGA65 by unmapping $A000-$BFFF rom region

gurcei:
While the existing c64's "unmap-basic.S" code made use of the c64's register $01 to unmap the C64 BASIC rom, the MEGA65 had an additional $D030 register to unmap rom regions related to the c65/mega65.

So I've borrowed "link.d" and "unmap-basic.S" from 'mos-platforms/c64' and tweaked them to suit the MEGA65.

Along the way, I repaired a build issue on Windows. Existing cmake was trying to download a windows ".zip" file and was getting a 404 "not found" error, as the file had been renamed to "llvm-mos-windows-main.7z" (7zip extension).


cmake/bootstrap-compiler.cmake
```diff
@@ -8,7 +8,7 @@ if(LLVM_MOS_BOOTSTRAP_COMPILER)
             CACHE STRING "The name of the llvm-mos release to download for precompiled tools")
   endif()
   if (WIN32)
-    set(extension zip)
+    set(extension 7z)
```
mlund, line 11: Could this change break compatibility with older Windows versions?

mlund (approved):
Looks good to me!

cmake/bootstrap-compiler.cmake
gurcei: I can share more of the backstory to provide context. While performing this build step under Windows:
```
cmake -DCMAKE_INSTALL_PREFIX=c:/src/llvm-mos -G "Ninja" ..
```
It resulted in this error:
```
CMake Error at llvm-mos-subbuild/llvm-mos-populate-prefix/src/llvm-mos-populate-stamp/download-llvm-mos-populate.cmake:170 (message):
  Each download failed!

    error: downloading 'https://github.com/llvm-mos/llvm-mos/releases/download/llvm-mos-windows-main/llvm-mos-windows-main.zip' failed
```

When I saw the that if failed to retrieved the .zip file from that url, I tried to visit the url in my browser and got the "404 Not Found" message.

However, within the `README.md` instructions, within the "Downloads" section, I noticed that the Windows link actually pointed to a .7z file instead:

- https://github.com/llvm-mos/llvm-mos-sdk/releases/latest/download/llvm-mos-windows.7z

So my gut feeling is that the project may have released Windows versions as *.zip files in the past, but no longer does so, and now releases the Windows version in a *.7z file.

Aah, but if the question is moreso whether the releases should remain as *.zip and not switch to *.7z? I'm not sure, I don't know the backstory on that.

But at least from the perspective of anyone newly cloning the sdk and trying to build it under Windows, the fix above will at the very least save them some grief trying to figure out why it breaks here.

mysterymath: Yeah, this was just an oversight; we should always fetch this as a 7z. I switched things over to 7z due to the increasingly ridiculous size of the Windows images; through this and other efforts, we've managed to get them down to more reasonable levels.

mysterymath (approved):

PR #119, opened 2023-06-17
Add RPC8/e platform.

asiekierka:
The RPC8/e platform is the virtual computer provided by the once-popular Minecraft mod [RedPower 2](https://www.eloraam.com), as well as [an independent emulator by Bigfoot](http://bigfootinformatika.hu/65el02/). Seeing as llvm-mos already had provisions for supporting the 65EL02 as a target (most likely due to Bigfoot's 64TASS changes for supporting it making it into upstream 64TASS), I decided to implement it as a full platform, complete with support for all of its built-in peripherals.

`librpc8e` follows the coding convention I've used in my own toolchains in the past; if warranted, I'm willing to adjust this convention to be more LLVM-friendly.

Technically, the 65EL02 supports 65816-style 16-bit register widths, which the backend doesn't emit; but that has never stopped people from trying to use cc65 in the past, and - more importantly - the 65EL02 does not support or use a 24-bit address space.


mos-platform/rpc8e/crt0.S
```diff
@@ -0,0 +1,46 @@
+;
+; Copyright (c) 2023 Adrian "asie" Siekierka
```
mysterymath, line 2: As far as is possible, we've licensed everything in the compiler and SDK under Apache 2.0 with LLVM exceptions. There's a bit of MIT floating around in the SDK, but I'd like to get rid of it as soon as we can get llvm-libc up an running; having a simple licensing structure is a definite benefit.

Would it be acceptable to license these files under those terms?

mos-platform/rpc8e/librpc8e/include/rpc8e.h
```diff
@@ -0,0 +1,33 @@
...
+ *    in a product, an acknowledgment in the product documentation would be
+ *    appreciated but is not required.
+ *
+ * 2. Altered source versions must be plainly marked as such, and must not be
+ *    misrepresented as being the original software.
+ *
+ * 3. This notice may not be removed or altered from any source distribution.
+ */
+
+#pragma once
```
mysterymath, line 23: Prefer old-fashioned include guards; `#pragma once` ties in a poorly-specified, compiler-specific notion of file identity that is more difficult to reason about than a textual identifier. (Soft and hard links, bind mounting, etc make "is this the same file" a complex question).

mos-platform/rpc8e/crt0.S
asiekierka: With the exception of `librpc8e/src/display.c` (where some snippets of code are derived from RPC8/e support work by L. Adamson done for cc65 back in 2012), it is acceptable to license all files under these terms.

asiekierka: I have relicensed the crt0 code to match; what should be done about the non-crt0 library?

mos-platform/rpc8e/librpc8e/include/rpc8e.h
asiekierka: Done.

mos-platform/rpc8e/crt0.S
mysterymath: Actually, the cc65 license is compatible with Apache 2.0; we're free to relicense code under that license so long as we follow its terms, which are just that we need an attribution notice for the source of that code, and a notice saying that they were modified from their original versions. We've done that with quite a few files already; most of the SDK headers are based on the cc65 equivalents.

asiekierka: Done. This leaves the examples, where I feel a CC0 license (or similar, such as 0BSD) would be justifiable to encourage reuse of such sample code; but I can relicense those too, of course.

mysterymath: Eh, that's fine, the risk there is low. The only thing you really gain with Apache2+LLVM over that is patent use and clear contributor relicensing clauses, but I wouldn't expect the examples there to be subject to continuous development.

I've mostly just tried to keep the licensing structure of the project in line with the implicit expectations folks should have with a project like this: it shouldn't be possible to start swinging patent or copyright law around to kill the project or annoy contributers, and there shouldn't be any onerous burdens on those using the output of the compiler, including any code compiled into the SDK. The LLVM exceptions help with that a lot, since they explicitly free compiled versions of the SDK from needing to bundle the license, which would otherwise be quite onerous for retro gamedev.

In practice, it probably doesn't matter; you'd have to be pretty bored and angry to sue a bunch of retro hackers. But, it hasn't been *too* difficult to structure things well "above board"; so far so good, anyway!

mysterymath (approved):

mos-platform/rpc8e/crt0.S
asiekierka: I'm glad we're on the same page with regards to choice of license! I've chosen the zlib license for my toolchain projects in the past, but I might migrate to Apache 2.0+LLVM myself

PR #121, opened 2023-06-18
[PCE] Add new platform.

asiekierka:
The PCE platform corresponds to the [NEC PC Engine](https://en.wikipedia.org/wiki/TurboGrafx-16), also known as the TurboGrafx-16 in the US. It is an interesting target for llvm-mos due to its particularly fast (7.16MHz) 6502 CPU, as well as large (up to 1MB; up to 8.5MB with extra mapper logic - not implemented here) ROM size and flexible banking.

The pull request will be marked as ready once I've collected enough feedback on it. Feel free to comment on what's already here, though - I'd be more than happy to receive feedback!

TODO (for this PR):

- [x] [Provide a solution for banking.](https://twitter.com/cbrzeszczot/status/1670432897625366529)
- [x] Provide a solution for interrupt handling, modeled after the NES target.
- [x] Add just enough VDC/VCE functions to allow bringing up a screen. Maybe.

Prerequisite PRs:

- [x] Fix HuC6280-related assembler bugs in llvm-mos. => waiting on https://github.com/llvm-mos/llvm-mos/pull/304
- [x] Add custom output format bounds support. => waiting on https://github.com/llvm-mos/llvm-mos/pull/306 (required for >8KB fixed bank)
- [x] [Add llvm-mlb support.](https://twitter.com/cbrzeszczot/status/1670497866987479041) => waiting on https://github.com/llvm-mos/llvm-mos/pull/305

Documentation: https://llvm-mos.org/wiki/PCE_target


mysterymath:
> The PCE platform corresponds to the [NEC PC Engine](https://en.wikipedia.org/wiki/TurboGrafx-16), also known as the TurboGrafx-16 in the US. It is an interesting target for llvm-mos due to its particularly fast (7.16MHz) 6502 CPU, as well as large (up to 1MB; up to 8.5MB with extra mapper logic - not implemented here) ROM size and flexible banking.
> 
> The pull request will be marked as ready once I've collected enough feedback on it. Feel free to comment on what's already here, though - I'd be more than happy to receive feedback!
> 
> TODO (for this PR):
> 
> * [x]  [Provide a solution for banking.](https://twitter.com/cbrzeszczot/status/1670432897625366529)
> * [x]  Provide a solution for interrupt handling, modeled after the NES target.
> * [ ]  Add just enough VDC/VCE functions to allow bringing up a screen. Maybe.
> 
> Prerequisite PRs:
> 
> * [ ]  Fix HuC6280-related assembler bugs in llvm-mos. => waiting on [Fix ELF relocation and relaxation for multi-operand 65C02/HuC6280 opcodes. llvm-mos#304](https://github.com/llvm-mos/llvm-mos/pull/304)
> * [ ]  Add custom output format bounds support. => waiting on [Add bounds support to lld custom output formats. llvm-mos#306](https://github.com/llvm-mos/llvm-mos/pull/306) (required for >8KB fixed bank)
> * [ ]  [Add llvm-mlb support.](https://twitter.com/cbrzeszczot/status/1670497866987479041) => waiting on [Add PCE platform support to llvm-mlb. llvm-mos#305](https://github.com/llvm-mos/llvm-mos/pull/305)
> 
> LMA setup:
> 
> Bits	Description
> 31 - 24	Group. 0x00 => base PCE address space; 0x01 => PCE card RAM address space (CD-ROM, Populous); 0x02 - 0x11 => "Street Fighter 2" mapper address space
> 23 - 16	Bank. For group 0x00: 0x00 - 0x7F => ROM, 0xF7 => CD backup RAM, 0xF8 => RAM, 0xF9 - 0xFB => SuperGrafx RAM, 0xFF => I/O. For group 0x01: 0x40 - 0x5F => RAM (Populous), 0x68 - 0x7F => RAM (Super System Card), 0x80 - 0x87 => RAM (PC Engine CD). For groups 0x02-0x11: 0x40 - 0x7F => ROM.
> 15 - 0	VMA

Is there a quick write-up I could read on the semantics of the vbank system? It would be useful to have documentation of the form avaiable on the Wiki for the NES targets; this would make it much easier to review this, and should help folks build a mental model of how the target works.

E.g. see https://llvm-mos.org/wiki/NES_targets

mos-platform/pce/libpce/src/memory.S
```diff
@@ -0,0 +1,124 @@
...
+; a subroutine in the imaginary register area and jump to it.
+
+.include "imag.inc"
+
+.global pce_memop
+.global pce_rmemcpy
+.global pce_memmove
+.global pce_memset
+
+    .section .text.pce_rmemcpy, "ax", @progbits
```
mysterymath, line 19: For the little self-modifying assembly I've written in llvm-mos; I just wrote an assembly language routine, placed it in the data segment, and added labels to each instruction to be modified. Then, another routine can use those labels to adjust the routine, then simply jump to it. This minimizes the overhead of calling the routine, since only the portions that change between calls are overwritten, and the result is very readable, being actual assembly.

asiekierka: Six of the eight bytes are caller-provided; as such, I believe the performance gain from doing so would be marginal (especially as the routine is laid out in such a way that - at least for `memcpy()` - two of the six caller-provided bytes are already in the right place at the moment of call), while the RAM use would increase by eight bytes (and we only have 8KB!).

The "ideal" way I'd like to resolve this is to emit the block copy instructions (TII/TDD) at the compilation stage if all the operands are known, minimizing the usage of these routines.

asiekierka:
Initial wiki page: https://llvm-mos.org/wiki/PCE_target

asiekierka:
Okay, I'm done with this PR for now.

Well, almost - currently there's no way to wait for a VBlank, but I'm out of ideas on how to add it, mostly because (a) I don't know how to best integrate C code into a NES-target-style interrupt system - it doesn't seem to be documented anywhere, (b) there's a bit of consideration that has to go into a "proper" IRQ handler, and I'm thinking on how to tackle it best; worst-case, I'll just make a second PR for that.

mos-platform/pce/libpce/include/pce/vdc.h
```diff
@@ -0,0 +1,283 @@
...
+
+/**
+ * @brief Set the VDC's copy method to copy every word.
+ */
+void pce_vdc_set_copy_word(void);
+/**
+ * @brief Set the VDC's copy method to copy every row.
+ * Alias for @ref pce_vdc_set_copy_word .
+ */
+#define pce_vdc_set_copy_row() pce_vdc_set_copy_word()
```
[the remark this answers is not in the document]
mysterymath: Actually I can never keep the semantics of C inline in my head; I've no idea whether it'll actually de-dup static inline versions of these. I've been in the habit of just making these kinds of things real functions in the .c file; it's pretty reasonable to assume that if you want fast code on llvm-mos you're doing LTO.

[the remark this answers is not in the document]
asiekierka: Admittedly, I've been in the habit of doing weird things like this because most previous platforms I worked with either lacked LTO or had buggy LTO.

Do you think it's a good idea to just standardize on non-static/non-inline functions across the library, then?

mos-platform/pce/libpce/src/memory.S
```diff
@@ -0,0 +1,124 @@
...
+pce_memmove:
+    ldy __rc4
+    cpy __rc2
+    bcc pce_rmemcpy
+    ldy __rc5
+    cpy __rc3
+    bcc pce_rmemcpy
+    bra pce_memcpy
+
+    .section .text.pce_memop, "ax", @progbits
```
mysterymath, line 62: There's a little choice point here: bunding pce_memop and __pce_memop_inner together like this makes them a unit for linker GC; accordingly the 14 byte prefix will always be included, even if never used. You could instead break them into separate sections and have a bra between them; this would add an extra instruction to pce_memop, but it would save those 14 bytes in the case that only e.g. pce_memmove and friends are used, which seems likely.

mos-platform/pce/libpce/src/psg.c
```diff
@@ -0,0 +1,24 @@
...
+#include <stdint.h>
+#include "pce/hardware.h"
+#include "pce/memory.h"
+
+__attribute__((alias("pce_psg_reset")))
+void __pce_psg_init(void);
+
+void pce_psg_reset(void) {
+    *IO_PSG_VOLUME = 0;
+	for (uint8_t i = 0; i < 6; i++) {
```
mysterymath, line 18: nit: indentation

mos-platform/pce/link.ld
```diff
@@ -0,0 +1,72 @@
...
+
+	/* Used by zero-bss routines. */
+	HIDDEN(__zero_bss_start = __bss_start + 1);
+	HIDDEN(__zero_bss_size = __bss_size >= 2 ? __bss_size - 1 : 1);
+	HIDDEN(__zero_zp_bss_start = __zp_bss_start + 1);
+	HIDDEN(__zero_zp_bss_size = __zp_bss_size >= 2 ? __zp_bss_size - 1 : 1);
+
+	.vector : {
+		. = 0xfff6 - 35;
+		KEEP(*(.early_start))
```
mysterymath, line 69: May be wise to break out early_start into its own section and ASSERT that SIZEOF that section is 35. Otherwise, this would produce weird error about an overflow of a rom section, if it produces one at all.

mysterymath (approved):

mos-platform/pce/libpce/src/memory.S
asiekierka: The only situation in which this would help is if only `pce_memset` is used; all the other.

Besides, these functions need a bit of a rewrite anyway - studying Elmer's notes reveals that, contrary to my belief, HuC block move instructions *do* stall interrupts, making them infeasible as a general-case memcpy replacement. Might be best to simply drop them (with the exception of pce_memop) from the initial release, which is what I'm going to do.

asiekierka:
Done. I've also removed the following features (in separate commits, in case we want to revert those in the future):

* pce_memcpy/pce_memmove/pce_memset - useful as they are, they currently block interrupts for the duration of executing the block transfer; this is good enough for a low-level function like `pce_memop`, but might need to be reconsidered for higher-level wrappers that replace C memcpy/memmove/memset.
* the virtual bank system - as per a lengthy discussion on Discord, it's better to add a feature in the future than to remove one, and virtual banks have the following two problems:
  * they are likely to be wholly superseded by automatic bank layouting,
  * `mos24bank(vbank_symbol)` wouldn't always return the expected physical bank - if the virtual bank was over 8KB, and the symbol was more than 8KB into the virtual bank, it would still return the first physical bank of the virtual bank, not the one the symbol actually resides in. This is a minor issue, but frustrating if far pointers were to be implemented.

mysterymath (approved):

Issue #113, opened 2023-04-28
Support cx16-specific kernal calls?


bcampbell:
The cx16 target supports the standard C64 kernal calls, but there are also a whole bunch of useful cx16-specific ones.
Is there any appetite for supporting shims for those calls to be directly available from C too?

It's a bit more complicated than the C64 calls - there are a bunch of more complicated functions and my (slightly sketchy) reading of things is that the cx16 kernal uses a different register convention, so a little shim is needed to move the params to the right places...

As an example, here's my shim for [`memory_decompress`](https://github.com/commanderx16/x16-docs/blob/master/X16%20Reference%20-%2004%20-%20KERNAL.md#function-name-memory_decompress).
I think I based it more or less on the pattern established in the `mos-platform/commodore)/cbm_k_*` functions.
But my definition and usage of the `__rcN` virtual registers is pretty dodgy!
Suggestions on how to do it better would be most welcome ;-)

```
; uint8_t* cx16_k_memory_decompress(uint8_t* src, uint8_t* dest);
;
; Kernel signature: void memory_decompress(word input: r0, inout word output: r1);
; Purpose: Decompress an LZSA2 block
; Call address: $FEED

__memory_decompress = $FEED

__rc2 = $04
__rc3 = $05
__rc4 = $06
__rc5 = $07

.global cx16_k_memory_decompress
cx16_k_memory_decompress:
;	lda mos8(__rc5) ; desthi
;	pha
;	lda mos8(__rc4) ; destlo
;	pha
;	lda mos8(__rc3) ; srchi
;	pha
;	lda mos8(__rc2)	; srclo
	lda __rc5 ; desthi
	pha
	lda __rc4 ; destlo
	pha
	lda __rc3 ; srchi
	pha
	lda __rc2	; srclo

	sta $02	; srclo
	pla
	sta $03	; srchi
	pla
	sta $04	; destlo
	pla
	sta $05	; desthi

	jmp __memory_decompress
```

I've got a few other cx16 functions I'd like to implement, so if anyone's interested I'd be happy to build up at least some partial coverage... (most likely mouse, joystick. keyboard fns)

mysterymath:
IIRC, @XarkLabs was pretty far along on this. I think there were some suspected codegen issues... to @XarkLabs: is there anything I can do help get this landed?

XarkLabs:
[...]

PR #123, opened 2023-06-19
[CX16] Use non-contiguous imaginary register convention to prevent ov…

asiekierka:
…erlap with the X16 KERNAL imaginary registers.

This is a proposed solution for the problem raised in https://github.com/llvm-mos/llvm-mos-sdk/issues/113 . Note, though, that, quoting [the KERNAL specification](https://github.com/X16Community/x16-docs/blob/master/X16%20Reference%20-%2004%20-%20KERNAL.md), "the 16-bit ABI *generally* follows the following conventions". There are a few functions which claim to affect even the so-called "saved" registers, so these will still need stashing - in accordance with the claimed convention.

As a bonus, this PR fixes the size of the zero page user area - `__basic_zp_end` - to match [the specification's memory map](https://github.com/X16Community/x16-docs/blob/master/X16%20Reference%20-%2007%20-%20Memory%20Map.md).

Paging @XarkLabs @bcampbell for their thoughts - I have never worked with the Commander X16 myself.

### DIFF
--- a/mos-platform/commodore/commodore.ld
+++ b/mos-platform/commodore/commodore.ld
@@ -8,7 +8,7 @@
 /* Provide imaginary (zero page) registers in the BASIC area. */
 __rc0 = __basic_zp_start;
 INCLUDE imag-regs.ld
-ASSERT(__rc31 == (__rc0 + 0x1F), "Inconsistent zero page map.")
+/* No assertion - cx16 uses a non-contiguous imaginary register block larger than 32 bytes. */
 
 MEMORY { zp : ORIGIN = __rc31 + 1, LENGTH = __basic_zp_end - (__rc31 + 1) }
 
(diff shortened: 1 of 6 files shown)


bcampbell:
Eyeballing it, it looks pretty good.
I could imagine having the first chunk of cx16 registers shifted in relation to the cx16 registers could be confusing in some circumstances - rs1-rs4 => r0-r3 etc...
But I think it's only the kernal shims which would ever have to worry about it. Once they are there nobody else ever has to deal with that detail (other than wondering why on earth the mos-llvm registers aren't a contiguous chunk on cx16 :- )

I haven't been able to build it and try it out yet - I think it requires "-mcpu=mos65el02" which means I'd have to build llvm-mos too? (it fails on the prebuilt llvm-mos release I'm using).

mysterymath (approved):

bcampbell:
> I haven't been able to build it and try it out yet - I think it requires "-mcpu=mos65el02" which means I'd have to build llvm-mos too? (it fails on the prebuilt llvm-mos release I'm using).

Ahh! I should have been getting the pre-built releases from llvm-mos-sdk, not llvm-mos! Will try again when I get a chance.

PR #125, opened 2023-06-26
[CX16] Fix imaginary register layout.

asiekierka:
I got caller-saved and callee-saved registers mixed up for this part.

Leaving to @XarkLabs to verify that I'm not missing something else - it's not a platform I work with regularly, or at all...


mysterymath (approved):
Yeesh, I got this backwards while reviewing it too.

asiekierka:
Given Xark's `I think it is in pretty good shape, per that latest PR. 🙂`, I think this can be un-drafted.

mysterymath (approved):
Did another quick once-over w the llvm-mos and x16 docs open; lgtm to merge.

PR #127, opened 2023-06-29
[PCE] Add pce-cd target.

asiekierka:
The PC Engine CD is a tougher piece of hardware to support in LLVM-MOS than the PC Engine itself. While banking was already a well-established concept, this one involves creating CDs! As such, a special tool called `pce-mkcd` was developed, which combines ELFs and binary files into a .ISO file, allowing symbol lookup to link in sector offsets of other binary files and even symbols from other ELFs.

TODO:

- [x] Implement System Card BIOS wrapper functions.
  - Some functions remain, but I'd say it's about 75% implemented. These can be done in a follow-up PR.
- [x] Make it possible to "opt out" of certain BIOS functionality (reclaiming Zero Page/RAM space).
- [x] Document everything on the wiki.
- [x] Add support for `@list.txt`-style arguments to pce-mkcd.


utils/pce-mkcd/pce-mkcd.cc
```diff
@@ -0,0 +1,668 @@
...
+          uint32_t r_file_offset = addr_to_file_offset(r_addr);
+          uint32_t r_sym_index = ELF32_R_SYM(int_at(elf_data, rela + ELF32_RELA_INFO));
+          uint32_t r_sym_type = ELF32_R_TYPE(int_at(elf_data, rela + ELF32_RELA_INFO));
+          uint32_t r_sym = symtab_offset + r_sym_index * symtab_size;
+          std::string r_sym_name = strtab_name(int_at(elf_data, r_sym + ELF32_SYM_NAME));
+          uint32_t r_sym_value = int_at(elf_data, r_sym + ELF32_SYM_VALUE);
+          uint16_t r_sym_shndx = short_at(elf_data, r_sym + ELF32_SYM_SHNDX);
+
+          // All resolved symbols are relocated - look only for unresolved symbols.
+          if (starts_with(r_sym_name, cd_symbol_prefix)) {
```
mysterymath, line 418: Just thinking out loud, but it looks like this has a pretty hefty chunk of a linker in it. Have you experimented with ways to get the linker to do these relocations instead? Maybe something along the lines of having the toolchain emit a relocatable .o file as its "final" output using LLD `-r`, but with a linker script to assign the output sections, just not their addresses. Then analyzing that object to extract sizes, and running the linker again to relocate everything, including the ones dependent on the CD layout.

Not sure whether something like that would work, but if it could be made to, it should help make this easier to maintain; ideally something as small as a relatively thin wrapper around llvm command-line tools; it might not need a detailed understanding of ELF at all.

asiekierka: I wouldn't say it's that hefty; compared to `elftocpm65`, the main difference is that it can also parse symbol names, and that I covered the sizes of all relocations "just in case".

I haven't experimented with getting the linker to do it much for the following reasons:

* I think `-r` would make it much more complicated, if not impossible, to assign things to absolute locations within a bank using a custom link script - binaries for the PC Engine CD are not unlike ROM images, just loaded at specific offsets in the 21-bit banked address space, and with the current approach, PCE CD link scripts are mostly similar in structure to existing PC Engine link scripts.
* How would the ability to link to another object's symbol be resolved?
  * Link them all together: this would cause annoying address space and symbol name overlaps. They are distinct programs, after all, and - more frustratingly - you may want to link to program B's `_start` address from program A (say, a game intro -> main game situation). An alternate solution of adding headers in front of every binary would open up the question of where to store them during loading - the CD has 2KB sectors, so a header would need to be padded to 2KB, while the unexpanded PCE CD has 72KB RAM total.
  * Provide a giant list of symbol definitions: you'd still need to generate addresses somehow before the final linking stage, in that case!
  
Do you see a way to accomplish this, while keeping a similar level of flexibility, which I don't?

mysterymath: > Do you see a way to accomplish this, while keeping a similar level of flexibility, which I don't?

Nope, I just wanted to prod for alternatives. Having this depend on the full list of relocs means we'd need to keep this in sync with the linker and MC layer, which adds a bit of maintenance overhead to otherwise unrelated code. If there's any way around that that isn't too much worse, that would be desirable, but if there's not, there's not.

Elfcpm65 was a bit of a special case, in that it supports a set of relocs, and those are the relocs it supports. It's converting from one relocatable object file format to another, while this is acting a bit more.... linkey

asiekierka: It doesn't have to depend on the full list of relocations; I just saw no reason not to add all of the ones supported so far (with the exception of special-case ones like `PCREL` or `ASCIIZ`), given the cost is one line per relocation.

mysterymath: That's cool then, if the intent is to support a subset of available relocs for those outstanding symbols. I think the most likely future relocs would be related to PIC when and if iget around to it, so that all should be fairly separable

asiekierka: I've added a comment, as to make it less confusing to future onlookers.

mos-platform/pce-cd/cd-ram-banked-sections.ld
```diff
@@ -0,0 +1,137 @@
...
+__ram_bank129_bank = 129;
+__ram_bank130_bank = 130;
+__ram_bank131_bank = 131;
+__ram_bank132_bank = 132;
+__ram_bank133_bank = 133;
+__ram_bank134_bank = 134;
+__ram_bank135_bank = 135;
+
+MEMORY {
+  ram_bank104 (rwx) : ORIGIN = (23592960 + __ram_bank104), LENGTH = 0x2000
```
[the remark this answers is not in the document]
asiekierka: Done - also done it retroactively for the regular `pce` target.

mos-platform/pce-cd/libpce/src/cd/bios.c
```diff
@@ -0,0 +1,640 @@
...
+uint16_t pce_cdb_version(void) {
+    uint8_t x, y;
+    __attribute__((leaf)) asm volatile(
+        ASM_BIOSCALL(0x1E) : "=x"(x), "=y"(y) : : "a", "p"
+    );
+    return (x << 8) | y;
+}
+
+void pce_cdb_irq_set(uint8_t type, void (*irq_handler)(void)) {
+    uint16_t irq_handler_address = (uint16_t) &irq_handler;
```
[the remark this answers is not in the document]
asiekierka: Fixed, thank you!

utils/pce-mkcd/pce-mkcd.cc
```diff
@@ -0,0 +1,765 @@
...
+  va_start(ap, msg);
+  fprintf(stderr, "Error: ");
+  vfprintf(stderr, msg, ap);
+  fprintf(stderr, "\n");
+  exit(result);
+}
+
+// Vector utilities.
+
+static uint8_t byte_at(std::vector<char> vec, size_t offset) {
```
mysterymath, line 90: const std::vector<char>&, here and everywhere; otherwise the standard requires this to memcpy the whole vector, which seems like it'll commonly be fairly large.

utils/pce-mkcd/pce-mkcd.cc
```diff
@@ -0,0 +1,765 @@
...
+
+static std::vector<unsigned char> vector_at(std::vector<char> vec, size_t offset, size_t size) {
+  std::vector<unsigned char> sub(size);
+  memcpy(&sub[0], &vec[offset], std::min(size, vec.size() - offset));
+  return sub;
+}
+
+// String utilities.
+
+static bool starts_with(std::string haystack, std::string needle, bool case_sensitive = true) {
```
mysterymath, line 110: Similarly const & here, although it doesn't matter quite so much.

utils/pce-mkcd/pce-mkcd.cc
```diff
@@ -0,0 +1,765 @@
...
+                case R_MOS_ADDR16_HI: width = 1; value >>= 8; break;
+                case R_MOS_ADDR24: width = 3; break;
+                case R_MOS_ADDR24_BANK: width = 1; value >>= 16; break;
+                case R_MOS_ADDR24_SEGMENT: width = 2; break;
+                case R_MOS_ADDR24_SEGMENT_LO: width = 1; break;
+                case R_MOS_ADDR24_SEGMENT_HI: width = 1; value >>= 8; break;
+                case R_MOS_FK_DATA_4: width = 4; break;
+                case R_MOS_FK_DATA_8: width = 8; break;
+                case R_MOS_IMM16: width = 2; break;
+                default:
```
mysterymath, line 511: "Unsupported" rather than "Unknown"; this gives a hint about the subsetting.

mysterymath (approved):

utils/pce-mkcd/pce-mkcd.cc
asiekierka: Done! Thank you; I haven't worked with C++ anywhere near enough.

mysterymath: Haha, or too much, depending on your perspective ;)

PR #131, opened 2023-07-03
Cx16 kernal support

XarkLabs:
These are the latest changes for the X16 platform.  Everything seems to be working quite well in the testing so far (with new register layout etc.).


mos-platform/cx16/cx16.h
```diff
@@ -296,5 +305,164 @@ struct __emul {
...
+    int x, y;
+} mouse_pos_t; /* cx16_k_mouse_get() */
+
+typedef struct
+{
+    unsigned char mode, columns, rows;
+} screen_mode_info_t;   /* cx16_k_screen_mode_get() */
+
+/* Kernal-level functions */
+void cx16_k_clock_get_date_time(cx16_date_time_t *datetime_ptr);
```
mysterymath, line 358: All of these that are implemented in assembly and cannot call back into the user program or anything in the platform libraries should be marked as `__attribute__((leaf))` (i.e., they're leaf calls from the POV of the compiler). Otherwise, the compiler has no idea what they might call, so it will have to assume anything that calls them can recurse, which forces use of a cc65-style soft stack.

It should be that all or almost all of these are leaves, but that's something to quickly verify for each wrapper.

mos-platform/commodore/cbm_k_load.s
```diff
@@ -3,11 +3,13 @@
 .text
 
 ;
-; unsigned int cbm_k_load(unsigned char flag, unsigned addr);
+; void *cbm_k_load(unsigned char flag, void *startaddr);
```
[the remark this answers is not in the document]
XarkLabs: Good catch.

mos-platform/cx16/cx16.h
XarkLabs: As discussed, doing these leaf changes in later PR (since they also affect cbm etc.).

mysterymath (approved):
LGTM!

PR #133, opened 2023-07-07
add __attribute__((leaf)) for cbm and cx16

### LINKED ISSUE #132: Missing `__attribute__((leaf))` in Commodore targets.
Various commodore-family KERNAL calls are not properly marked with `__attribute__((leaf))`, which forces the compiler to pessimize what they might call. They'll usually call nothing; most of them are leaves WRT the C environment. Accrordingly, those implemented in assembly should be marked as leaf in the header, and those implemented in C should have any internal declarations marked as leaf. (This only for functions that are actual leaves WRT the LTO link.)

XarkLabs:
This is to resolve #132 adding left attribute to functions.


mysterymath (approved):
Thanks for taking the time!

PR #137, opened 2023-07-16
cx16 CBM file access example (plus strcat, strncat)

XarkLabs:
This is a CBM file access test and example for cx16.  It also adds a basic C strcat and strncat implementation (that should be cross platform).

### DIFF
--- a/examples/cx16/k_cbm_file_test.c
+++ b/examples/cx16/k_cbm_file_test.c
@@ -10,6 +10,14 @@
 #include <cbm.h>  // Commodore kernal functions
 #include <cx16.h> // X16 specific kernal functions
 
+// NOTE: I believe the file access routines used here are
+//       portable to other CBM machines, however, this
+//       example does make use of a few minor X16 features
+//       mostly dealing with ASCII use 80 column formatting
+//       and some X16 specific control codes.
+//       Currently it has only been tested on X16 (with
+//       llvm-mos and cc65 [messed up screen I/O on cc65]).
+
 uint8_t device_num = 8; // CBM device number (8 = main disk)
 uint8_t cbm_error;      // CBM error code
 char cbm_status[64];    // CBM device status string
@@ -176,6 +184,7 @@ char read_buffer[256];
 
 void show_status(void)
 {
+    // X16: control codes defines
     printf("%s=%c%d \"%s\"%c\n", cbm_error < 20 ? "success" : "failure", cbm_error < 20 ? CH_WHITE : CH_LIGHTRED, cbm_error, cbm_status, CH_WHITE);
 }
 
@@ -183,7 +192,7 @@ int main(void)
 {
     int16_t len;
 
-    putchar(CH_FONT_ISO); // use ISO character set mode (for upper/lower)
+    putchar(CH_FONT_ISO); // X16: use ISO character set mode (for upper/lower)
     printf("Simple test of CBM file access routines.\n\n");
     cbm_get_status();
     printf("Initial device 8 status: \"%s\"\n\n", cbm_status);
--- a/examples/cx16/k_cbm_file_test.c
+++ b/examples/cx16/k_cbm_file_test.c
@@ -1,0 +1,282 @@
+// llvm-mos-sdk cx16 kernel file access tests
+//
+// vim: set et ts=4 sw=4
+
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+
+#include <cbm.h>  // Commodore kernal functions
+#include <cx16.h> // X16 specific kernal functions
+
+uint8_t device_num = 8; // CBM device number (8 = main disk)
+uint8_t cbm_error;      // CBM error code
+char cbm_status[64];    // CBM device status string
+char cbm_tempstr[128];  // CBM temp for device commands
+
+uint8_t cbm_get_status(void);
+uint8_t cbm_device_cmd(const char *cmd);
+uint8_t cbm_delete_file(const char *fn);
+uint8_t cbm_rename_file(const char *old_fn, const char *new_fn);
+uint8_t cbm_open_file(uint8_t lfn, const char *fn, const char rw);
+uint8_t cbm_close_file(uint8_t lfn);
+long cbm_file_read(uint8_t lfn, void *buf, size_t maxsize);
+long cbm_file_write(uint8_t lfn, const void *buf, size_t size);
+
+// read device status channel, returns cbm_error and sets cbm_status
+uint8_t cbm_get_status(void)
+{
+    uint8_t len = 0;
+    cbm_k_setlfs(15, device_num, 15);
+    cbm_k_setnam("");
+    cbm_k_open();
+    cbm_k_chkin(15);
+    while (len < sizeof(cbm_status) && !cbm_k_readst())
+    {
+        uint8_t c = cbm_k_chrin();
+        if (c == '\r')
+        {
+            break;
+        }
+        cbm_status[len++] = c;
+    }
+    cbm_status[len] = 0;
+    cbm_k_clrch();
+    cbm_k_close(15);
+    cbm_error = ((cbm_status[0] - '0') * 10) + (cbm_status[1] - '0');
+    if (cbm_error < 20) // quash trivial errors
+    {
+        cbm_error = 0;
+    }
+
+    return cbm_error;
+}
+
+// send device command
+uint8_t cbm_device_cmd(const char *cmd)
+{
+    cbm_k_setlfs(15, device_num, 15);
+    cbm_k_setnam(cmd);
+    cbm_k_open();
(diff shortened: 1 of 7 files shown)


mos-platform/common/c/string.c
```diff
@@ -120,3 +120,19 @@ __attribute__((weak)) char *strtok(char *string, const char *separators) {
   savestring = s2;
   return s1;
 }
+
+__attribute__((weak)) char *strcat(char *restrict s1, const char *restrict s2) {
+  unsigned int len = strlen(s1);
```
mlund, line 125: `unsigned int` -> `size_t`

mos-platform/common/c/string.c
```diff
@@ -120,3 +120,19 @@ __attribute__((weak)) char *strtok(char *string, const char *separators) {
...
+  strcpy(s1 + len, s2);
+}
+
+__attribute__((weak)) char *strncat(char *restrict s1, const char *restrict s2,
+                                    size_t n) {
+  unsigned int len = strlen(s1);
```
mlund, line 131: `unsigned int` -> `size_t`

mos-platform/common/include/string.h
```diff
@@ -32,6 +32,9 @@ size_t strspn(const char* string, const char* in);
 char* strpbrk(const char* string, const char* brk);
 char* strtok(char* string, const char* separators);
 
+char *strcat(char * __restrict__ s1, const char * __restrict__ s2);
```
mlund, line 35: It's not used elsewhere, but perhaps Doxygen style comments could be added? Some IDE's can pick them up and display them as popup help.

examples/cx16/k_cbm_file_test.c
```diff
@@ -0,0 +1,282 @@
+// llvm-mos-sdk cx16 kernel file access tests
+//
+// vim: set et ts=4 sw=4
```
mlund, line 3: Add licence infor and author

examples/cx16/k_cbm_file_test.c
```diff
@@ -0,0 +1,282 @@
...
+
+#include <stddef.h>
```
mlund, line 5: `k_cbm_file_test.c` does not seem to specific to x16, but would work on any commodore target. Could it be lifted to a `commodore/` folder? And what does `k_` signify in the cx16 tests?

mlund (approved):
Very nice to have a test of CBM disk features, useful for other cbm-like targets. Please see single-line comments.

mos-platform/common/include/string.h
mlund: Perhaps copy in `.clang-format` and `.clang-tidy` from the llvm-mos rep? Styles vary throughout `string.h`, e.g. `char*` vs `char *`.

examples/cx16/k_cbm_file_test.c
mlund: Instead of the vim comment, rely on `.clang-format` (see above comment) or perhaps `.editorconfig` (https://editorconfig.org).

XarkLabs: None of the other examples I peeked at have license or author.  I prefer the options per file in easy to edit or see comment (vim: stuff).  clang-format is different, I wish you had a .clang-format (like the lllvm one) in the llvm-mos-sdk tree (and it would automatically get used).

examples/cx16/k_cbm_file_test.c
XarkLabs: This is the prefix on the functions being exercised. I believe it was inherited from the cc65 "API" and stands for KERNAL [sic].  It _almost_ works on other CBM platforms, but not fully.  Also, I don't believe there is a "generic" commodore example folder (or any other CBM flavored examples - other than the ones I have added for X16).

mos-platform/common/c/string.c
XarkLabs: My bad on these.

mos-platform/common/include/string.h
asiekierka: It is used elsewhere in the repo - I've added them to my PC Engine and RPC8/e libraries. As such, there is precedent.

XarkLabs: Sure, I can get the llvm clang-format (I've think I have done it previously, but I get pristine or work on a different machine and it goes away). Why not add the format you desire to the tree so it (mostly) defaults?  I didn't feel comfortable reformatting the whole file (but I would love not having to think much about formatting - I need some kind of clang format to write code and show me where I need to balance braces 🙂).  I looked a bit into .clang-tidy, but not clear how to integrate into my git flow (I am not great with git at this point).

mlund: `.clang-format` should probably be added and applied in a separate PR.

examples/cx16/k_cbm_file_test.c
mlund: Thanks, I suspected it was `kernal` but it seems a little odd not knowing the history. I think it would be useful to add a `commodore/` folder and a test that works across all CBM targets. Perhaps this is for another PR, but we should probably avoid duplicate code. Could you add a comment pointing out where X16 specific stuff is?

mos-platform/common/include/string.h
XarkLabs: Okay, I verified string.c did match up llvm clang-format (at least didn't change on reformat).  However, now I remember, I really dislike the llvm clang-format brace dangling on the end thing, so do I really have to use it on the X16 examples I create?😅  It wasn't mentioned on the other examples I previously submitted.  But, if that is policy, then so be it. I can understand keeping it all consistent and in lib code and files I didn't create I have been trying to be a "code chameleon" (after some previous newbie mistakes).🦎

XarkLabs: Also, I was not really aware of Doxygen use.  Cool, going forward I can look into decorating some things to make that (more) useful.

examples/cx16/k_cbm_file_test.c
XarkLabs: The main X16 issues I am aware of (e.g., from compiling with cc65) are mostly due to all the "normal ASCII" use all over.  Other issues are the X16 specific control characters used.  That said, I suspect pretty trivial to "port" (and perhaps make more ugly with 40 columns in all uppercase or add ASCII to PETSCII conversion [or maybe some exists - but I am not a fan of cc65 style "char mangling"]).  I'll add a few portability notes.

mos-platform/common/include/string.h
mlund: I guess it's just a matter of taste and probably something Daniel should decide on at a project level. In any case, it can/should be changed at a later stage so no worries.

examples/cx16/k_cbm_file_test.c
mlund: > None of the other examples I peeked at have license or author.

You're right there's no file-level license info elsewhere in the SDK and I guess the LICENSE file at top-level suffices. The llvm project often adds a header to source files, though.

asiekierka: File-level licensing information is a good practice to minimize ambiguity.

XarkLabs: I would be fine with adding some info to the cx16 examples (just following other examples).  Maybe a few other ones can get updated and I will copy the "style" and language for the cx16 ones.

mysterymath: There's already a requirement to include a copy of the Apache license with any distribution of the sources; per-file license just provide a backup if this isn't done, since it's separately a license violation to remove them, and that takes an additional action, rather than just say forgetting to include LICENSE.

Accordingly, I haven't been in the habit for requiring Apache license headers outside of files where there already license headers; there the callthrough to the LICENSE file should always be made explicit, since otherwise the existing headers would seem to supercede LICENSE. The llvm-mos repo is also a special case, since it intended to look as much as possible like the surrounding files.

That being said, I'm not opposed to having license headers, or further discussion along this line. If folks would for instance feel put out by having these files appear somewhere else without attribution, that would be a good reason to add headers. I just don't know of a compelling reason on my end.

> I wish you had a .clang-format (like the lllvm one) in the llvm-mos-sdk tree (and it would automatically get used).
I'll second this; I'd prefer to see a copy of the `.clang-format` in llvm-mos added to this repo. We should do this in a separate step with an SDK-wide refactoring though; it cuts the git history, and we should only do that once. Accordingly, this should be out of scope for this PR.

mos-platform/common/include/string.h
mysterymath: I've been really lax on this, but now that it's a question, enforcing the llvm style is probably the most straightforward non-terrible thing we can do. It has machine support via clang-format, and it's relatively easy to git clang-format a commit before it goes in.

Agree that this should be separate; we should do a refactoring for style once and only once; otherwise we'll cut the git history more than is strictly necessary.

PR #147, opened 2023-08-03
common: add atoi(), atol()

asiekierka:


mlund:
Thanks for this very quick reply! It looks good; the only thing I can think of is if there could be some overlap reduction with the internal version here? https://github.com/llvm-mos/llvm-mos-sdk/blob/666870714f215acd892642557cd1a4a2b6943af9/mos-platform/common/c/printf.c#L187

asiekierka:
Yes, but `printf.c`'s `_atoi` is not compliant with the C standard. The `atoi` in `stdlib` is a superset.

mos-platform/common/c/atoi.cc
```diff
@@ -0,0 +1,22 @@
...
+	T result = 0;
+	bool sign = false;
+
+	while (isspace(*str)) { str++; }
+
+	if (*str == '+') { str++; }
+	else if (*str == '-') { sign = true; str++; }
+
+	while (*str >= '0' && *str <= '9') {
+		result = (result * 10) + (*(str++));
```
mysterymath, line 14: Shouldn't this be `(*(str++) - '0')`?

mysterymath (approved):

mos-platform/common/c/atoi.cc
asiekierka: Nice catch!

Made sure to *actually* verify if it works this time...

PR #149, opened 2023-08-06
Port to Ben Eater's 6502 Breadboard Computer

rweather:
This is a port of LLVM-MOS-SDK to Ben Eater's 6502 Breadboard Computer.  The port is very basic, befitting a very basic 6502 computer design.  Some interesting features of this port:

- .text and .rodata are placed in ROM (actually EEPROM), .data and .bss are placed in RAM.
- Serial console using the W65C51N ACIA for both sending and receiving at 19200 bps N-8-1.
- Serial reception is interrupt-based with a 256-byte buffer to avoid missing characters.  Serial transmission busy-waits for characters to be written due to a bug in the W65C51N that prevents interrupt-based transmission from working.
- System millisecond tick counter function millis() using the T2 timer in the W65C22 VIA.  Also has a delay() function for simple delay loops.
- The user's program can provide irq() and nmi() interrupt handlers if they need special handling beyond the built-in serial and T2 interrupt handlers.

The linker script was hacked up based on the one for "atari8-stdcart" to get the infrastructure needed for ROM'ing the .text and .rodata segments.  If there is a better linker script to copy for ROM'ing, then let me know and I'll fix it.

Not done yet, but may be useful in the future:

- Support for the LCD interface in the standard Ben Eater design.
- API to make it nicer to use the W65C22 VIA, similar to cx16's _6552.h include file.  I wasn't sure whether copying this file from the cx16 directory was the correct thing to do or not, so I didn't include that in this patch.

More information is in mos-platform/eater/README.md.


mos-platform/eater/link.ld
```diff
@@ -0,0 +1,49 @@
...
+       INCLUDE text-sections.ld
+  } >rom AT>rom
+  .rodata : { INCLUDE rodata-sections.ld } >rom AT>rom
+  .data : { INCLUDE data-sections.ld } >ram AT>rom
+  __data_load_start = LOADADDR(.data) - LOADADDR(.rodata) + ADDR(.rodata);
+  __data_size = SIZEOF(.data);
+  .zp.data : { INCLUDE zp-data-sections.ld } >zp AT>rom
+  __zp_data_load_start = LOADADDR(.zp.data) - LOADADDR(.rodata) + ADDR(.rodata);
+  __zp_data_size = SIZEOF(.zp.data);
+  .data_initializers ADDR(.rodata) + SIZEOF(.rodata) + __data_size +
```
mysterymath, line 29: The `.data_initializers` section was a symptom of the older split LMA/VMA approach that I'd used on banked ROM targets, but it ended up being unnecessary. I had reworked the NES targets so this isn't needed, but had forgotten to go back and re-do `atari-stdcart`; unlucky.

In any case, it's unnecessary here, since the ROM has both its LMA and VMA in the same memory region: `rom`. Accordingly, you can just get rid of `.data_initializers`.

mysterymath (approved):

rweather:
Thanks for the feedback.  I have made the requested changes.

mysterymath:
Thanks for the new target!

PR #153, opened 2023-08-19
Add hardware accelerated multiplication and division on MEGA65

mlund:
This replaces the software multiplication and integer division in the C runtime with accelerated variants that use the MEGA65 32-bit hardware math registers. Multiplication is 1 cycle; division up to 20 cycles. The `*` and `/` operators as well as `div()` automatically use the new functions for 8, 16, and 32 bit (un)signed integers.

- [x] `umul()`, `udiv()`, `udivmod` replaced for 8, 16, and 32 bit
- [x] 64-bit integer manipulation is not accelerated and uses the generic software implementation. This could be sped up by re-writing as a series of 32-bit operations.
- [x] Add `check-math.cc` example to the MEGA65 examples directory for verifying the output
- [ ] Expose `common/crt/divmod.h` internally and use as fallback in mega65 target
- [x] Make mul/div atomic


mysterymath (commented):
I'm afraid these routines aren't reentrant. If a multiplication or division is interrupted, and the interrupt handler performs another multiplication or division, it would clobber the one being established.

The simlest way I've found to fix this is to create a static boolean wasInterrupted flag that gets set to false at the beginning of the procedure and set to true at the successful end. Then, attempting the operation is done in a loop, repeating until wasInterrupted is false. In a nested chain of interrupts, the topmost will complete, informing the next one down that it was interrupted, which will retry, informing the next one down it was interrupted, etc., until the full chain is complete. For best granularity; this would need two bools: one for mul, and one for div.


Let me know if you can think of or know of a better way to do this. There's almost certainly a stock way to deal with this, but I don't know it and wasn't able to find a well-known name for the problem.

mos-platform/mega65/crt/divmod.h
```diff
@@ -0,0 +1,244 @@
...
+
+// Relatively straigtforward implementation of long division in C. Not
+// particularly tuned for performance, but clear.
+
+template <typename T> static inline T udiv(T a, T b) {
+  if constexpr (sizeof(T) <= 4) {
+    return udiv_m65(a, b);
+  }
+
+  if (!b || b > a)
```
mysterymath, line 30: Instead of copying this, just call this udiv_m65, and call the above udiv_m65_hardware or something. Then, this can just call through to the original udiv rather than duplicating it. Same for multiplication.

asiekierka:
Given it takes relatively few cycles, wouldn't it be possible to just push processor flags, disable interrupts, then pop processor flags at the end?

mysterymath:
For processors where it's possible to completely disable interrupts, that works (and it's the usual way of handling this), but the issue persists for NMI handlers. For usual functions, saying "you can't call this in an NMI handler" can be a sensible limitation, but that's a harder sell for compiler-generared libcalls, since the user doesn't have direct control over when they're generated.

rweather:
Maybe an incrementing "protect" variable rather than a TRUE/FALSE flag to avoid race conditions in checking and setting the boolean:

```
redo:
   INC protect
   LDA protect
   STA temp
   ... ; do the operation
   LDA temp
   CMP protect
   BNE redo
```

If the "protect" variable changes while doing the operation, then the result is unpredictable and the process repeats.  Note that it isn't necessary to block interrupts, as long as the "INC protect" instruction is atomic.

The above assumes that there are less than 256 levels of interruption before control returns.  This might be an issue with pre-emptive multitasking and a system with more than 256 processes and interrupt handlers active at once.

It would be useful to have atomic builtins in llvm-mos that do this safely ;e.g. https://github.com/llvm-mos/llvm-mos/issues/138

mysterymath:
A boolean is safe so long as it's volatile and guards the region. See the MMC1 example:

https://github.com/llvm-mos/llvm-mos-sdk/blob/2d0451cfae0cde5c98250370f57bf6d36c8ad41a/mos-platform/nes-mmc1/bank.c#L72

If a reentrant call interrupts the guarded region, _IN_PROGRESS is definitely set to false, causing a retry. In this case, the guarded region is immediately following the store to _IN_PROGRESS and ending at the reload, both of which are atomic.

mysterymath:
A boolean is safe so long as it's volatile and guards the region. See the MMC1 example:

https://github.com/llvm-mos/llvm-mos-sdk/blob/2d0451cfae0cde5c98250370f57bf6d36c8ad41a/mos-platform/nes-mmc1/bank.c#L72

If a reentrant call interrupts the guarded region, _IN_PROGRESS is definitely set to false, causing a retry. In this case, the guarded region is immediately following the store to _IN_PROGRESS and ending at the reload, both of which are atomic.

mos-platform/mega65/crt/divmod.h
mlund: Yes, that's a very good point - I initially removed the original implementation, but discovered that it's needed for 64-bit ints. Will fix.

mlund:
Thanks for the feedback - I'll work on a solution as suggested.

mos-platform/mega65/crt/divmod.h
mlund: I'm entirely sure what CMake magic to apply to expose the common crt to derived targets. I the header file should be available only during building and not part of the install target.

mysterymath: `target_include_directories` set to `PRIVATE` I think; these affect building of the SDK in CMake, but shouldn't affect the installed SDK.

mos-platform/mega65/crt/divmod.h
```diff
@@ -0,0 +1,244 @@
+#ifndef __SLOW_DIV
+
+#define DIVBUSY 0xD70F
+#define DIVOUT_FRAC 0xD768
+#define DIVOUT_WHOLE 0xD76C
+#define MULTINA 0xD770
```
mysterymath, line 6: Since we're in C++, `static auto &MULTINA = *(volatile unsigned long *)(0xD770);`, that should help make the results more readable.

mlund:
> A boolean is safe so long as it's volatile and guards the region. See the MMC1 example

I added the guard following the MMC1 example. Does it look reasonable? Next step is to expose `common/crt/divmod.h`.

mysterymath:
> > A boolean is safe so long as it's volatile and guards the region. See the MMC1 example
> 
> I added the guard following the MMC1 example. Does it look reasonable? Next step is to expose `common/crt/divmod.h`.

Yep, this looks right to me.

mlund:
This has been stale for a while and will close until I can get back to this.

PR #154, opened 2023-08-21
[mega65] Add MEGA65 headers for hardware registers

mlund:
This adds headers for MEGA65 MMIO in the spirit of CC65 and reused headers from C64 where possible (SID, CIA).
- [x] Header for VIC-IV with structured registers. Includes defines for bit masks generated from the MEGA65 core file `iomap.txt`.
- [x] Header for VIC-III with mask defines only.
- [x] `mega65.h` header with basic hardware layout inspired by KickC. It also adds math registers and hypervisor traps.
- [x] Add two MEGA65 C/C++ examples that use the new headers
- [x] Care taken to make headers compatible with CC65 and Calypsi


mysterymath (commented):
I'll have to look through this tonight, so just a general comment for now.

The style of these headers differs pretty heavily from the cc65-derived ones used for the other commodore targets. I'm led to understand that cc65 doesn't currently support the mega65, and it sounded like it might be difficult to add.

I'm not super familiar with mega65 development in practice, but does it seem likely that cc65 will gain support in the future? Whatever gets checked in here will likely be something we have to support for the forseeable future, so it wouldn't be super-great if we end up having two incompatible versions of these headers for mega65. Are these similar enough to how ASM programmers typically program the system? Is there good prior art for Mega65 C, etc. etc.?

Also, is the hardware for the mega65 a conservative extension the c64, at least logically? It seems like it would be nice if C code written against the c64 would compile for the mega65 if so; that seems like it would be reason enough to reframe these to extend the cc65 interfaces. This could be done piecemeal; for example, if the Mega65 SID doesn't actually add anything over the c64 one, it would seem more natural to reuse the cc65 definitions rather than introducing a new set.

mlund:
As far as I know there's no current development of a mega65 target for cc65. Everything I have seen is compiled in C64 mode. There is a fork somewhere, and I should take a closer look to see if there're useful headers. And yes, now checking the KickC SID header and it's silly to add defines to $d400 as there's actually four SID chips.

nobruinfo:
@mysterymath The Mega65 has two main modes as you already mentioned. With `GO64` it is possible to take all stuff already available for the real C64 but extensions are possible. The native Mega65 mode (so no `GO64` fired, a bit like on the C128) allows for bank switching, VIC-IV, DMA, Hypervisor access, SID and later OPL... and a lot of more stuff. The thing is, you can also do this in `GO64` mode! So it will be the now challenge to have some HAL for LLVM-MOS to start with.

What has been made for KickC is actually a good start. I'm already using it in my projects exactly like this. It will not be finished and yes, future folks will say as you and will want a different and then incompatible approach. On the other hand not having the includes and more stuff will not let new folks find it and not use LLVM at all.

mysterymath:
No worries; I definitely don't want to hold up a change like this too long to make the interfaces look *just so*, especially where there isn't any real prior art to be compatible with.

Accordingly, for this change, I'd recommend re-using the existing cc65 headers for the chips that made it into the mega65 verbatim; that way we'll preserve the property that's anything truly common between the commodore-family targets can be spelled the same way. We can then feel free to use the KickC headers verbatim for the mega65-specific or mega65-extended chips; it's of minimal use trying to create "cc65-style" versions of those headers, since there's no backwards compatibility concern, and if and when someone comes to make cc65 headers, they may end up incompatible anyway.

mlund:
As there doesn't seem to be any standard, it's a good chance to design one to our liking. As CC65 is used in the MEGA65 kernal it would make sense to reuse headers and stick to the same style. But perhaps the best place to do this is in mega65-libc and then eventually include that with the SDK? It has a cmake friendly build system and we could use `ExternalProject_Add()` do fetch and build a tagged version.

mysterymath:
> But perhaps the best place to do this is in mega65-libc and then eventually include that with the SDK? It has a cmake friendly build system and we could use `ExternalProject_Add()` to fetch and build a tagged version.

The SDK should definitely be as standalone as we can make it; it's CMake is already fairly convoluted, and adding out-of-tree dependencies only makes that worse. This is essentially a stand-in for a monorepo; I'd put everything in the llvm-mos repo if it wasn't already so impossibly big and it's CMake so crazy.

mlund:
> The SDK should definitely be as standalone as we can make it

Right, I also suspect that LGPL3 will be an issue, so let me work on the headers and then we can worry about where to put them. I will make them MIT/Apache licensed.

mlund:
OK, the headers are now much simplified and re-uses the SID and CIA headers from C64.

nobruinfo:
@mlund Thank you very much from the compiler/library user's side. This must have been days of highly concentrated and somewhat tedious work. But we're all going to profit from that. Over in Mega65 Discord there's already folks wanting to make stuff with it. I think your work will be pointing them, us all to the right direction.

mos-platform/mega65/mega65.h
```diff
@@ -0,0 +1,229 @@
...
+
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wgnu-anonymous-struct"
+#pragma clang diagnostic ignored "-Wnested-anon-types"
+#pragma clang diagnostic ignored "-Wgnu-binary-literal"
+#pragma clang diagnostic ignored "-Wfixed-enum-extension"
+#endif
+
+#include <_45E100.h>
```
hth313, line 47: As I understand it these underscore first files are commonly used in cc65 community. I never liked it and think that it is better to prefix it with a path, like `mega65/6526.h`, compared to `_6526.h`.
To me it seems clear that `mega65/6526.h` is a header that defines the 6526 chip to be used on the MEGA65.
The `_6526.h` leaves me in doubt whether the included header file works with the MEGA65, it may define things for a 6526, but if it provides a base symbol, will it work with the MEGA65, it is not clear.
I do not mind having all these `_6526.h` and whatever if there is a compatibility reason for it, but I would prefer the blessed interface would be "system/chip.h" and it can delegate as needed. 
I honestly do not know if it applies here, but at least it would be nice if the public API was this way.

mlund: The naming follows the existing style which is inherited from CC65 with the main argument of compatibility. My understanding is that `_` hints "private" and a user would only include `mega65.h` or `c64.h` which is unambiguous. For the specific case of `_6526.h` it resides in the C64 target and thus valid for multiple platforms. It might actually be better to elevate this and the vic chips to `commodore` as they are used by C64, MEGA65, and the currently unimplemented C128. I will leave that for another PR.

hth313: Ah, I see. Yeah, it kind of works with "private". I suppose the idea is to include `<mega65.h>` or `<c64.h>` if you are on C64. I suppose they could be considered to move in under `commodore` (or perhaps `mos` as the chip manufacturer?) at some point as you suggest.

mlund:
I'll wait a little while with these until mega65-libc licensing has been settled.

mlund:
Right, I think it's time to reopen this PR. The headers should be compatible with CC65 and Calypsi. They are independent of mega65 libc.

mysterymath (approved):
LGTM!

PR #180, opened 2023-09-04
Initial Action 53 (mapper 28) support

jroweboy:
Action 53 is a NES multicart mapper used often for game jam competitions because its both cheap to produce, and easy to target. While developers can use the existing CNROM or NROM targets to make an Action53 compatible game, it won't be able to take advantage of all of the banking features in the mapper such as CHR banking. Adding native A53 mapper support will hopefully entice more developers that join the NESDEV Game Jam to use LLVM-MOS.

NOTICE: I haven't had much battle testing with this as I just barely finished bootstrapping the code. I plan to fix any issues found with the banking implementation as I find them, but this is at least enough to generate a basic ROM that runs.


mos-platform/nes-action53/prg-rom-64.ld
```diff
@@ -0,0 +1,15 @@
+INCLUDE _prg-rom-banked.ld
+
+__prg_rom_size = 64;
+
+OUTPUT_FORMAT {
```
mysterymath, line 5: As far as I can tell, it isn't necessary to break out separate linker scripts for 32 and 64 PRG-ROM sizes; the symbol value setting should be enough. If `__prg_rom_size == 32`, the `prg_rom_1` and `prg_rom_2` sections will have length zero, so the `FULL()` statements below won't add any content for them. That would let the default linker script PROVIDE() a default value for `__prg_rom_size`, and the user could override it just by defining a value for that symbol, either in inline assembly or a `settings.s` file.

mysterymath (approved):

mysterymath:
Thanks for contributing this by the way! It's really heartening to see that someone could make sense of how the existing NES scripts were put together and extend it to a new mapper without me; I had hoped this was possible, but it was an open question whether it would actually happen. 

If you get the chance, would you mind adding a quick blurb to https://llvm-mos.org/wiki/NES_targets about what this supports and what the current limitations are?

jroweboy:
Rebased and updated to address the review comments. I'll add a blurb to the wiki about this now

PR #183, opened 2023-09-06
Fix a few Action 53 bugs

jroweboy:
Two bug fixes

* Add explicit NMI ordering for all of the NES target mappers with NMI callbacks. This fixes an issue where the neslib nmi isn't guaranteed to be last (which can cause all sorts of problems when it starts doing VRAM updates during the main gameplay)
* Fixes an issue where banking would not take effect because the PRG bank size wasn't set in the reset routine. The also defaults the mirroring mode to VERTICAL (but its easy enough for users to change)

### DIFF
--- a/mos-platform/nes-action53/_prg-rom-banked.ld
+++ b/mos-platform/nes-action53/_prg-rom-banked.ld
@@ -15,10 +15,10 @@ MEMORY {
   prg_rom_0     : ORIGIN = __prg_rom_0_lma,     LENGTH = __prg_rom_size >= 32  ? 0x4000 : 0
   prg_rom_1     : ORIGIN = __prg_rom_1_lma,     LENGTH = __prg_rom_size >= 64  ? 0x4000 : 0
   prg_rom_2     : ORIGIN = __prg_rom_2_lma,     LENGTH = __prg_rom_size >= 64  ? 0x4000 : 0
-  prg_rom_fixed : ORIGIN = __prg_rom_fixed_lma, LENGTH = 0x4000 - 0x30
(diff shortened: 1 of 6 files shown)


mos-platform/nes-action53/_prg-rom-banked.ld
```diff
@@ -15,10 +15,10 @@ MEMORY {
   prg_rom_0     : ORIGIN = __prg_rom_0_lma,     LENGTH = __prg_rom_size >= 32  ? 0x4000 : 0
   prg_rom_1     : ORIGIN = __prg_rom_1_lma,     LENGTH = __prg_rom_size >= 64  ? 0x4000 : 0
   prg_rom_2     : ORIGIN = __prg_rom_2_lma,     LENGTH = __prg_rom_size >= 64  ? 0x4000 : 0
-  prg_rom_fixed : ORIGIN = __prg_rom_fixed_lma, LENGTH = 0x4000 - 0x30
+  prg_rom_fixed : ORIGIN = __prg_rom_fixed_lma, LENGTH = 0x4000 - 0x32
 
   /* Reset stub for the fixed bank. */
-  reset_stub : ORIGIN = 0x10000 - 0x30, LENGTH = 0x10
```
mysterymath, line 21: No Action Required for this change, but does reset_stub actually have to be the very last thing in the address space, or can it just be assigned to prg_rom_fixed? The latter would save needing to adjust this if the stub changes sizes.

mysterymath (approved):

mos-platform/nes-action53/_prg-rom-banked.ld
jroweboy: It likely should, good point. This was a hold over from when i was first starting to debug the build and was testing out linker sections.

Issue #173, opened 2023-09-01
NES UNROM-512 mapper target


mysterymath:
This is reportedly one of the more popular mappers used for new NES development.

PR #191, opened 2023-09-10
Add NES UNROM-512 mapper support, mapper defines to subtargets.

asiekierka:
All sub-modes are supported, as is both CHR-ROM and CHR-RAM.

This also fixes "bus conflict" behaviour in the UNROM mapper implementation; well, it doesn't exactly *fix* it, it papers around it - as per the NESdev wiki, programmers should not rely on the "bitwise AND" behaviour of the bus during a conflict. It also marks the submapper in the NES 2.0 header as a non-bus-conflict one.

I'm not sure how to best fix it properly - given the major ROM usage increase required to do so (256 bytes, fewer if one could vary it by linker script), there's some discussion opportunity to be had here, IMO. Probably best for a new issue.


asiekierka:
Fixes https://github.com/llvm-mos/llvm-mos-sdk/issues/173

mos-platform/nes-unrom-512/bank.h
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * @brief Define this in a .c file to use 2-screen horizontal mirroring.
+ */
+#define UNROM_512_USE_2_SCREEN_HORIZONTAL \
+	asm(".globl __mirroring\n.globl __four_screen\n__mirroring = 0\n__four_screen = 0\n")
+
+/**
+ * @brief Define this in a .c file to use 2-screen vertical mirroring.
+ */
+#define UNROM_512_USE_2_SCREEN_VERTICAL \
```
mysterymath, line 22: Also, man this is a really nice pattern; we should have these for most settings in the SDK I think. NAR here; I made #193 for this.

mos-platform/nes-unrom-512/common.ld
```diff
@@ -0,0 +1,15 @@
+/* NES-UNROM-512 iNES file linker script. */
+
+INCLUDE nes.ld
+
+ASSERT(__chr_rom_size <= 32, "UNROM only supports up to 32 KiB of CHR-ROM.")
```
mysterymath, line 5: nit: UNROM-512

mos-platform/nes-unrom-512/prg-rom.ld
```diff
@@ -0,0 +1,44 @@
...
+  FULL(prg_rom_24)
+  FULL(prg_rom_25)
+  FULL(prg_rom_26)
+  FULL(prg_rom_27)
+  FULL(prg_rom_28)
+  FULL(prg_rom_29)
+  FULL(prg_rom_30)
+  FULL(prg_rom_fixed)
+  FULL(vectors)
+  FULL(chr_rom_0)
```
mysterymath, line 40: Just FULL(chr_rom); we don't encode the bank into the addresses, so that region always encompasses the whole CHR-ROM, regardless of target.

mos-platform/nes-unrom-512/bank.c
```diff
@@ -0,0 +1,44 @@
...
+// Licensed under the Apache License, Version 2.0 with LLVM Exceptions.
+// See https://github.com/llvm-mos/llvm-mos-sdk/blob/main/LICENSE for license
+// information.
+
+#include "bank.h"
+
+extern char _BANK;
+__attribute__((section(".zp.bss"))) char _BANK_SHADOW;
+
+__attribute__((leaf)) inline char get_bank_state(char mask) {
```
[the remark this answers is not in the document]
asiekierka: It's provided for completion alongside `set_bank_state` (now `set_bank_part`), which is somewhat less trivial.

mos-platform/nes-unrom-512/bank.c
```diff
@@ -0,0 +1,44 @@
...
+  return _BANK_SHADOW & mask;
+}
+
+__attribute__((leaf)) inline char set_bank_state(char mask, char value) {
```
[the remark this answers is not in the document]
asiekierka: I thought of `set_bank_part`, but `set_bank_bits` makes more sense.

mos-platform/nes-unrom-512/bank.s
```diff
@@ -18,7 +18,7 @@ banked_call:
 	sta __rc17 ; __rc17 = requested PRG bank
```
mysterymath, line 18: Ah, I hadn't noticed that the RC value is preserved across the call, right.

Since `__rc20` is caller-saved, the value on entry to `banked_call` needs to be saved and restored too, a la `lda __rc20; pha`.

asiekierka: Sorry! Should be fixed.

mysterymath: Aaand this clobbers the value on entry in A; the store to `__rc17` needs to happen first ;)

mysterymath (approved):
Awesome, LGTM!

Issue #193, opened 2023-09-11
Macros to set SDK linker script symbols


mysterymath:
Commonly symbol values are used to configure the linker scripts for a given platform. This is particularly widespread in the NES targets, since a given mapper target can usually support a wide variety of actual boards, each with slightly different physical hardware configurations. The usual syntax for this is to define the symbols either in an assembly file or in C, but the syntax for this is a little obtuse. We can make this easier by providing C preprocessor macros to set these values using module-level inline ASM. This was done in #191 for UNROM-512, but this issue is to generalize the pattern to the other targets.

asiekierka:
nit: The PC Engine (`pce`) target is a better - more complex/powerful - example of this macro setup in action.

asiekierka:
One question, though: how far should we go? Should we allow defining as many INES values from C as possible, not unlike the PC Engine target's approach to banking? (Think `MAPPER_PRG_ROM_KB(128);` and such.) Or should this just be about board configuration, like UNROM-512 currently?

asiekierka:
I don't think a generic "symbol setting" macro would save much, compared to the added complexity of nested C macros.

It could be a good idea to have a general `ines.h`, though.

PR #194, opened 2023-09-11
Add safe bus conflict handling.

### LINKED ISSUE #192: Better solution for NES mapper bus conflicts
Some mappers supported by llvm-mos (CNROM, some variants of UNROM, some variants of UNROM-512) feature bus conflicts:

* The CNROM mapper currently works around this by not providing a `set_chr_bank` routine, instead expecting the user to provide their own with the following pattern for each bank used:

```c
void set_bank(char b) {
  static const char rom_bytes[] = {0, [1] = 1, [255] = 255};
  POKE(&rom_bytes[b], b);
}
```

* The UNROM and UNROM-512 mappers (given https://github.com/llvm-mos/llvm-mos-sdk/pull/191) work around this by writing to a `$FF` ROM location every time, which often does - but [is not guaranteed to](https://forums.nesdev.org/viewtopic.php?p=109708#p109708) - lead to writing the correct value.

The correct solution is to find a way to make the CNROM mapper's solution more generally applicable - but can this be done without a 256-byte table forced on every such mapper?

* UNROM-512 has a documented non-bus-conflict variant, which would not require this table at all.
* UNROM and CNROM often only use a fraction of the table (N bytes are required for a ROM with N non-fixed PRG/CHR banks), they shouldn't need to pay the 256-byte "toll" every time.

Ideas/suggestions welcome!

asiekierka:
Fixes #192, or "How I Learned To Stop Worrying And Love The Linker".

Also adds SDK-side `get_chr_bank`/`set_chr_bank` helpers to CNROM, now that this is actually possible to do safely; as well as an UNROM 512 macro to enable a "self-flashing" board, now that this is no longer the default.

(Other INES macros will be added via #193)

I'm giving this one a longer write-up because I think the level of hacks in the PR necessitates an explanation.

## The Problem

The NES features [mapper bus conflicts](https://www.nesdev.org/wiki/Bus_conflict). Simpler mappers, based around discrete logic chips, do not prevent the ROM chip from being enabled when the CPU is writing to the cartridge's data bus - this creates a conflict, as both the CPU and ROM are outputting values on the same bus at the same time.

The correct way to work around this is to ensure that the value output by the CPU is the same as outputting the ROM. This is typically done by creating a byte array `A` where, for any given value `N`, `A[N] == N`; then, whenever one wants to write `N` to ROM, it is done to the address of `A[N]`.

However, this byte array would normally take up 256 bytes of the 16384 bytes typically present in the "fixed" ROM - this level of wastefulness is unacceptable. Many mappers only require a far smaller table; for example, an UNROM cartridge typically only has 16 banks, so the only values one would need to write to the ROM bus will be between 0 and 15. We'd like the LLVM-MOS SDK to be able to dynamically allocate, at link time, the right-sized array - of `M` bytes - based on the INES header configuration (the bank sizes, etc).

## The Answer

Naturally, it'd be nice to have an unified solution for all the affected mappers. By sheer coincidence, all the discrete mappers with bus conflicts which llvm-mos-sdk wishes to target have just one writable memory-mapped port, which is always reachable from the "fixed" ROM space. Nice! As such, a `rompoke.h` header is defined with the function `void rom_poke_safe(char value);`, which performs the `A[N] == N` operation.

Next, we need to create the dynamically allocated table.

The first idea was to generate one table - with a distinct section and symbol - per table size (from 0 to 256 inclusive - mappers with bus conflicts often have non-conflicting variants, which we can then avoid paying the table's size cost on completely), like so:

```
__rom_poke_table = big... ? __rom_poke_table_256 : (less_big ? __rom_poke_table_128 : ...)
```

Unfortunately, LLD will retain in ROM all symbols used in such an equation - which is worse than just having one table in the first place.

## The Terrible, Terrible Answer

Something we *can* dynamically control, though, is section sizes; we can also control the byte value used to fill such a section.

So, *technically speaking*, one could define 256 sections - one for each byte of `A` - and simply set the first `M` sections to a size of 1 byte, with the remaining `256 - M` sections set to a size of 0 bytes.

But who would ever do such a thing? Anyhow, here's the pull request.

## The Less Terrible But Actually No Good Answer

That's when I found that GNU's linker, which LLD claims drop-in compatibility with, states the following in its manual:

`If the fill expression is a simple hex number, ie. a string of hex digit starting with ‘0x’ and without a trailing ‘k’ or ‘M’, then an arbitrarily long sequence of hex digits can be used to specify the fill pattern; Leading zeros become part of the pattern too.`

This isn't that well-known - older versions of `ld` were limited to just two- or four-byte fill patterns.

Either way, that's it! I can have just one section which holds an arbitrary number of bytes based on a fill pattern, then set the section's size using `. = . + __rom_poke_table_size;`! This is great! It will certainly-

>ld.lld: error: [...]/mos-platform/nes/lib/rompoke-table.ld:4: malformed number: 0x000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f202122232425262728292a2b2c2d2e2f303132333435363738393a3b3c3d3e3f404142434445464748494a4b4c4d4e4f505152535455565758595a5b5c5d5e5f606162636465666768696a6b6c6d6e6f707172737475767778797a7b7c7d7e7f808182838485868788898a8b8c8d8e8f909192939495969798999a9b9c9d9e9fa0a1a2a3a4a5a6a7a8a9aaabacadaeafb0b1b2b3b4b5b6b7b8b9babbbcbdbebfc0c1c2c3c4c5c6c7c8c9cacbcccdcecfd0d1d2d3d4d5d6d7d8d9dadbdcdddedfe0e1e2e3e4e5e6e7e8e9eaebecedeeeff0f1f2f3f4f5f6f7f8f9fafbfcfdfeff

... If this works on the GNU linker (I checked with binutils 2.39), does this technically constitute an LLVM bug?

## The Summary

The PR is a terrible hack, but it does get the job done and it seemingly passes all the tests. Please be merciful.


mos-platform/nes/rompoke/misc/generate-rompoke-mapper-linkscript.lua
```diff
@@ -0,0 +1,152 @@
+-- Copyright (c) 2023 Adrian "asie" Siekierka
+--
+-- Licensed under the Apache License, Version 2.0 with LLVM Exceptions,
+-- See https://github.com/llvm-mos/llvm-mos-sdk/blob/main/LICENSE for license
+-- information.
+
+local CONFIG_MAP = {
```
mysterymath, line 7: It looks like if we go this way, adding new mappers with bank conflicts would practically need to involve modifying this script. Accordingly, this is a major part of the porting interface for the NES, so it needs a lot of cleanup, additional documentation, or both, such that it's possible to relatively quickly get a sense of how to do so. At least on my end, I've only formed a very, vague picture of how this works after studying it for 5 minutes or so.

mos-platform/nes-unrom-512/_rompoke.ld
```diff
@@ -0,0 +1,2 @@
+/* Automatically generated by generate-rompoke-mapper-linkscript.lua. */
```
mysterymath, line 1: Part of the generated output should be a human-readable description of the logic that this reduces to. This could just be hard-coded per target; that should also help make the intent for each target easier to read in the generator script.

mos-platform/nes/rompoke/rompoke-table.ld
```diff
@@ -0,0 +1,1028 @@
+/* Automatically generated by generate-rompoke-table-linkscript.lua. */
+SECTIONS {
+  .rom_poke_table_0 : {
+    __rom_poke_table = .;
+    FILL(0x00000000)
+    . = . + (__rom_poke_table_size > 0 ? 1 : 0);
+  } >c_readonly
+  .rom_poke_table_1 : {
+    FILL(0x01010101)
```
mysterymath, line 9: Does `FILL(0x01)` not work? If not, add a comment to specify this; the fact that these are 32-bits each is very suspicious and could use a rationale.

mos-platform/nes/rompoke/rompoke-table.ld
```diff
@@ -0,0 +1,1028 @@
+/* Automatically generated by generate-rompoke-table-linkscript.lua. */
+SECTIONS {
```
mysterymath, line 2: A little generated comment here to describe how each block works and the purpose would go a long way towards readability.

Also, this is a very nice hack; it depends only on well-trodden paths, and it thus seems likely to be maintainable.

mos-platform/nes/rompoke/rompoke-table.ld
asiekierka: Works just fine; it was a leftover.

mos-platform/nes/rompoke/misc/generate-rompoke-mapper-linkscript.lua
asiekierka: Should be taken care of, now.

mos-platform/nes-unrom-512/_rompoke.ld
asiekierka: Should be taken care of likewise.

mos-platform/nes/rompoke/rompoke-table.ld
asiekierka: The comment should be taken care of now.

mysterymath (approved):
Awesome, LGTM!

Issue #185, opened 2023-09-07
Struct of Array C++ library


mysterymath:
Using the struct of arrays pattern requires a constrained way of programming to maintain efficiency. It's unlikely that this optimization would pay off in arbitrary C++ outside of very restricted scenarios (e.g., a function-local array). It's also unlikely that the user experience of relying on a compiler optimization for this would be very good; there would be a lot of implicit, unwritten rules, and violating any of them would kill performance. Accordingly, providing a C++ library for this would give most of the benefit with a minimum of effort, and it would provide explicit source-level documentation that the pattern is being used, as well as help guard against certain type of misuse.

johnwbyrd:
Possible inspiration, though it seems to be married to STL types: https://github.com/crosetto/SoAvsAoS

PR #197, opened 2023-09-12
Add C++ struct-of-arrays library and example

mysterymath:
This adds two new types to the SDK under `soa.h`:
  - `soa::Array<T, s>` is an array of `T` of size `s`, but where the k'th byte of each `T` are stored contiguously.
  - `soa::Ptr<T>` is a pointer to an element of `soa::Array`.

The include-eable `soa-struct.h` specializes `soa::Ptr<T>` for a struct type, including its members as distinct `soa::Ptr<T>` instances that point to the members.

Using this library allows global arrays of structs to be accessed using the absolute indexed addressing mode if it's always known at compile-time what portion of the struct is being accessed. This is generally true if there are no pointers into the array; in that common case, using this library can produce a huge speedup.

See the included example for more details; the library is quite expansive in practice, so the examples walk through what you can and can't do with it.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -25,6 +25,8 @@ if(HOSTED)
   install_example(hello-putchar)
   add_executable(init-functions init-functions.cc)
   install_example(init-functions)
+  add_executable(struct-of-arrays struct-of-arrays.cc)
+  install_example(struct-of-arrays)
 endif()
 
 if(PLATFORM MATCHES ^nes-)
--- a/examples/struct-of-arrays.cc
+++ b/examples/struct-of-arrays.cc
@@ -2,12 +2,10 @@
 #include <soa.h>
 #include <string.h>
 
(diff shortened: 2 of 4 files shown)


mysterymath:
@asiekierka , @jroweboy , @johnwbyrd

mysterymath:
Closes #185

johnwbyrd:
This is a good and useful library.  When convenient, just because I suspect that SoA type transformations will seem alien to the old-school 6502 community and to Joe Programmer, you may wish to put in a readme.md file that describes why it's important and useful for generating fast code on the 6502.

mysterymath:
> This is a good and useful library. When convenient, just because I suspect that SoA type transformations will seem alien to the old-school 6502 community and to Joe Programmer, you may wish to put in a readme.md file that describes why it's important and useful for generating fast code on the 6502.

That's a good idea; I've modified the doc comment for the main class to include a broader introduction to the idea and why one would do this. That helps motivate everything about the library's syntax and semantics, so it's important that it's understood going in.

Issue #195, opened 2023-09-11
Audit NES shadow registers for non-reentrancy


mysterymath:
In cases where a memory location is shadowing a MMIO register, a NMI interrupt may occur between the set of the MMIO register and the set of the shadow, and that interrupt may make an incorrect determination of the MMIO state by reading the shadow register. We should analyze the SDK so far for such issues, see if there's a straightforward way to address them, and if not, explicitly call out in which circumstances it's guaranteed accurate to call those accessors.

asiekierka:
Here's some loose thoughts:

The typical shadowing routine for a MMIO register on a one-register mapper (UNROM, CNROM, etc.) would be (written with ASM this time):

```
tax
sta _BANK_SHADOW
sta __rom_poke_table, x
```

In this situation, there's only one point during which an NMI can fire where the shadowed register no longer matches the real register: the write between `_BANK_SHADOW` and `rom_poke_table`.

The easy way to resolve that is to always write the shadow to the ROM at the beginning of NMI:

```
lda _BANK_SHADOW
tax
sta __rom_poke_table, x
```

This has the nice property of unlocking two new functionalities:

- the ability to write to the ROM without writing to _BANK_SHADOW, creating a bank change that will automatically undo itself at the end of NMI, but invalidating getters,
- the ability to write to _BANK_SHADOW without writing to the ROM, creating a delayed bank change that will only apply at the end of NMI. Combined with re-applying _BANK_SHADOW at the beginning of NMI, this allows for "change CHR bank at vertical blank" handlers.

To borrow terms from the MMC1 bank implementation, I'm going to refer to the former as a "split".

Re-applying _BANK_SHADOW at the beginning of NMI also means that getters and setters will work correctly *inside* an NMI. The only case where they won't work correctly is "splits" - this should be appropriately documented; however, CHR banks don't tend to be paged in/out the same way a PRG bank would across calls, which makes me believe this won't be much of a problem in real-world use.

asiekierka:
This will work fine enough for those discrete mappers whose register corresponds to only one value (CNROM, UNROM). For UNROM-512, though, one needs to rethink it: we don't want a situation in which, say, writing _BANK_SHADOW to the register on `set_prg_bank` also changes the CHR bank.

mysterymath:
[...]

PR #198, opened 2023-09-13
NES: Rework CNROM, UNROM and UNROM-512 to properly handle NMIs.

asiekierka:
Fixes https://github.com/llvm-mos/llvm-mos-sdk/issues/195 for these three mappers. I *think* this leaves only MMC3 as the remaining mapper without NMI handling (MMC1 and Action53 already had their own; NROM does not require it).

I've implemented my own approach to handling NMIs:

* _BANK_SHADOW is always written to before the mapper's MMIO register itself;
* _BANK_SHADOW is written to MMIO at an early stage of NMI, allowing developer-side consistency for NMI code written in C once the AXY calling convention is added.

For CNROM and UNROM-512, I've extended this with additional functionality:

* The graphical part of the register (CHR bank, mirroring) is now deliberately separated between "shadow" (written to the MMIO register on NMI or on user request) and the MMIO register.
* Writing to the "shadow" without writing to the MMIO register is called a "swap" - this will only apply the result on the next frame. This allows preparing a CHR bank change without affecting the currently drawn frame.
* Writing to the MMIO register without writing to the "shadow" is called a "split" - this will only apply the result *until* the next frame. This allows doing consistent mid-frame effects.

As this may not be immediately intuitive to someone tinkering with the code in the future, I have added a test for these two mappers which automatically validates it.

This required removing `get_chr_bank` on CNROM - well, it didn't *strictly* require it, but then it would require two RAM bytes instead of one; I expect the functionality to be sufficiently unnecessary as to not be worth the added cost. Likewise, the public API of UNROM-512 has changed; as neither API has yet shipped in a production build of the SDK, this is currently not a breaking change.

### DIFF
--- a/mos-platform/nes-cnrom/bank.c
+++ b/mos-platform/nes-cnrom/bank.c
@@ -8,10 +8,37 @@
 
 __attribute__((section(".zp.bss"))) char _BANK_SHADOW;
 
-__attribute__((leaf)) char get_chr_bank(void) {
-  return _BANK_SHADOW;
+__attribute__((leaf)) void set_chr_bank(char value) {
+  asm volatile("" ::: "memory");
+  _BANK_SHADOW = value;
+  asm volatile("" ::: "memory");
+  rom_poke_safe(value);
+  asm volatile("" ::: "memory");
 }
 
-__attribute__((leaf)) void set_chr_bank(char value) {
-  rom_poke_safe(_BANK_SHADOW = value);
+__attribute__((leaf)) void swap_chr_bank(char value) {
+  asm volatile("" ::: "memory");
+  _BANK_SHADOW = value;
+  asm volatile("" ::: "memory");
 }
+
+__attribute__((leaf)) void split_chr_bank(char value) {
+  asm volatile("" ::: "memory");
+  rom_poke_safe(value);
+  asm volatile("" ::: "memory");
+}
+
+/**
+ * Pass _BANK_SHADOW to the physical bank value at the start of NMI. This
+ * prevents an inconsistency if the NMI is fired between _BANK_SHADOW is set
+ * and the ROM being written to. Doing so at the *start* of NMI allows using
+ * banked_call and the same "switch to new; run code; switch back to old"
+ * code pattern inside NMI handlers.
+ *
+ * TODO: Adapt to C when the AXY caller convention is present, allowing the
+ * compiler to optimize unused shadow variables out.
+ */
+asm(".section .nmi.10,\"axR\",@progbits\n"
+	  "lda _BANK_SHADOW\n"
+	  "tay\n"
+	  "sta __rom_poke_table,y\n");
--- a/mos-platform/nes-cnrom/bank.c
+++ b/mos-platform/nes-cnrom/bank.c
@@ -6,26 +6,19 @@
 #include <rompoke.h>
 #include "bank.h"
 
-__attribute__((section(".zp.bss"))) char _BANK_SHADOW;
+__attribute__((section(".zp.bss"))) volatile char _BANK_SHADOW;
 
 __attribute__((leaf)) void set_chr_bank(char value) {
-  asm volatile("" ::: "memory");
   _BANK_SHADOW = value;
-  asm volatile("" ::: "memory");
   rom_poke_safe(value);
-  asm volatile("" ::: "memory");
 }
 
 __attribute__((leaf)) void swap_chr_bank(char value) {
-  asm volatile("" ::: "memory");
   _BANK_SHADOW = value;
-  asm volatile("" ::: "memory");
 }
 
 __attribute__((leaf)) void split_chr_bank(char value) {
-  asm volatile("" ::: "memory");
   rom_poke_safe(value);
-  asm volatile("" ::: "memory");
 }
 
 /**
--- a/mos-platform/nes-cnrom/bank.c
+++ b/mos-platform/nes-cnrom/bank.c
@@ -6,15 +6,15 @@
 #include <rompoke.h>
 #include "bank.h"
 
-__attribute__((section(".zp.bss"))) volatile char _BANK_SHADOW;
+__attribute__((section(".zp.bss"))) volatile char _BANK_NEXT;
 
 __attribute__((leaf)) void set_chr_bank(char value) {
-  _BANK_SHADOW = value;
+  _BANK_NEXT = value;
   rom_poke_safe(value);
 }
 
 __attribute__((leaf)) void swap_chr_bank(char value) {
-  _BANK_SHADOW = value;
+  _BANK_NEXT = value;
 }
 
 __attribute__((leaf)) void split_chr_bank(char value) {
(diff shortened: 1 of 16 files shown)


mysterymath:
Having a threefold split between set, swap, and split seems a bit extra. Most commonly one would want to change the CHR for the next frame, uncommonly one may want to immediately change CHR for some graphical effect. The two are separable; one may or may not want the immediate chr bank split to set the next frame's CHR. But, if one did, one could take the additional action to set the next frame's CHR.

As for setting the next frame's CHR, there are two places where this might be done: inside VBLANK and outside. Outside, the right way to do this is to set a flag to be picked up inside VBLANK. Inside, the right way is to immediately and summarily set CHR. So, it seems like the isolate bits of functionality are:
  1. Setting the CHR bank applied automatically at start of NMI (MMC1's `set_chr_bank`), and
  2. Immediately setting the CHR bank (MMC1's `split_chr_bank`).
 
Both the NMI handler should set the shadow register and the actual bank.

So, what of interrupts? Well, (1) is atomic; it either applies or it doesn't; no problems there. But if (2) were interrupted, it was either by VBLANK NMI or another source. In the case of VBLANK NMI, the split just didn't make it; ideally, it should have no effect whatsoever, and the NMI handler should see the value set by (1). In the case of any other interrrupt, well, it could see an old value of the shadow register for the current bank; that's the original bug. In that case, it seems prudent to disable interrupts for the duration of the write; that way shadow updates can be kept atomic.

asiekierka:
Perhaps it'd help to get more eyeballs on the PR?

mysterymath:
Yeah, paging @jroweboy , if you get the chance.

mos-platform/nes-cnrom/bank.c
```diff
@@ -8,10 +8,37 @@
 
 __attribute__((section(".zp.bss"))) char _BANK_SHADOW;
```
mysterymath, line 9: This should always be a real shadow of the register; CNROM shouldn't be the only mapper without this. The other mappers have another byte to deal with this, and one should really be spent here too.

mysterymath: This should be volatile, similarly for the other shadows. They're written in regular contexts and read in interrupt contexts; that's classic volatile.

asiekierka: >This should always be a real shadow of the register

UNROM also uses only one byte; it just so happens that it doesn't have any CHR/video-related functionality. UNROM-512 uses two bytes, but that's because it has *both* CHR/video-related *and* PRG-related mapper configuration functionality, so it'd require two bytes regardless.

Is there a developer-side reason to allow reading back the CNROM shadow for CHR? It's not the kind of thing where you'd typically have a trampoline/"push-pop" pattern for development, like with doing things with PRG data... In addition, having a separate real shadow requires more cycles spent on upkeeping them (as now NMI can get fired between writing to shadow A and shadow B, as well as shadow B and the actual MMIO register).

Overall, I'm not quite convinced.

>This should be volatile, similarly for the other shadows. They're written in regular contexts and read in interrupt contexts; that's classic volatile.

Volatile will force the bank shadow to be present even if the code never reads from it; it is possible to construct a hypothetical NES game's code in such a way that the bank shadow would be unnecessary, thus making MMIO register writes atomic and speeding up the whole ordeal a bit. The reason I opted for memory barriers instead is to allow the compiler to optimize `_BANK_SHADOW` out entirely, but that is currently not possible as the NMI handler cannot be expressed with C code alone.

mysterymath: Another EDIT: My original reply was substantially quite incorrect; I had misremembered why I added these.

> > This should always be a real shadow of the register
> Is there a developer-side reason to allow reading back the CNROM shadow for CHR? It's not the kind of thing where you'd typically have a trampoline/"push-pop" pattern for development, like with doing things with PRG data... In addition, having a separate real shadow requires more cycles spent on upkeeping them (as now NMI can get fired between writing to shadow A and shadow B, as well as shadow B and the actual MMIO register).

I had originally thought that I had copied these over from nesdoug, but I went back to the original sources, and it looks like they were inventions of mine. The reason for their inclusion was then certainly because they seemed simple to include and for orthogonality with PRG-related shadows, but the closer examination done here reveals that they're not actually very simple to maintain.

There is still an orthogonality argument; if we pull out CHR-ROM getters, we should do it across mappers and take the ABI hit. Increasingly, I think that's what we should do; there's nearly always a good reason to maintain careful PRG state shadows (`banked_call`, but the case is harder to make for CHR, and it's something that a user can take care of themselves to their desired degree of fidelity. Library code is usually worse than user code, since we have to defensively deal with situations that the user can know never happen, so I'd be in favor of ripping all this out.
 
> Volatile will force the bank shadow to be present even if the code never reads from it; it is possible to construct a hypothetical NES game's code in such a way that the bank shadow would be unnecessary (NMI using fixed bank only + careful PRG bank arrangement), thus making MMIO register writes atomic and speeding up the whole ordeal a bit. The reason I opted for memory barriers instead is to allow the compiler to optimize `_BANK_SHADOW` out entirely, but that is currently not possible as the NMI handler cannot be expressed with C code alone; even then, I don't think the compiler would allow modeling what I'm trying to achieve here, so might be best to just move back to volatile.

Lost the original comment on this one, but the gist was this: the memory barriers are quite a bit harder to read than volatile would be, and it doesn't appear we get any gain from them today. When we can know for sure that there's a design that can use them to provide tangible benefit, that's a good time to put that in, but the best way to know something like that is to have an implementation of it in-hand. Taking the opposite policy tends to cause strange code to accumulate in projects like this, and it's difficult to root it out once it's forgotten whether or not its quirks are load-bearing.

mos-platform/nes-cnrom/bank.c
```diff
@@ -6,12 +6,32 @@
 #include <rompoke.h>
 #include "bank.h"
 
-__attribute__((section(".zp.bss"))) char _BANK_SHADOW;
```
mysterymath, line 9: This should definitely be called `_BANK_NEXT`, since it entirely controls the bank as of next NMI, and it has no relation to the current bank state.

mos-platform/nes-cnrom/bank.c
mysterymath: I think pulling out the CHR getters is fairly separate from this review, so we can table that discussion for this one and introduce a disparity between the mappers. The set/swap/split API seems worth including regardless. I'd also like to see it backported to MMC1/MMC3; we need to figure out the right way forward there, trading off nesdoug tutorial compatibility and internal SDK consistency.

mos-platform/nes-cnrom/bank.c
asiekierka: Done.

mos-platform/nes-cnrom/bank.c
asiekierka: I think pulling out the CHR getters for now is a good idea; it's less breaking to add an API some time later than to remove an API some time later.

mysterymath (approved):

Issue #174, opened 2023-09-01
NES GTROM mapper target


mysterymath:
This is reportedly one of the more popular mappers used for new NES development.

asiekierka:
Depends on https://github.com/llvm-mos/llvm-mos-sdk/issues/156

PR #199, opened 2023-09-14
NES: Add GTROM mapper.

asiekierka:
Fixes https://github.com/llvm-mos/llvm-mos-sdk/issues/174 .

This introduces an approach to handling NES mappers utilizing 32K PRG-ROM banks, inspired by @mysterymath 's work on MMC1 32K banking. In short, a faux "fixed" section is placed at the beginning of each bank, with the bank-specific contents following after. Thanks to clever use of the custom output format's origin/length functionality, the "fixed" bank is dynamically sized to fit the user's code and data exactly.

An important caveat is that this approach does not currently provide a good answer for DPCM support on this mapper (as the fixed section begins at $8000, not $C000).


asiekierka:
Added DPCM support as discussed on Discord. This should be everything, now.

mos-platform/nes-gtrom/_prg-rom-banked.ld
```diff
@@ -0,0 +1,259 @@
...
+  prg_rom_14 : ORIGIN = __prg_rom_14_lma, LENGTH = __prg_rom_size >= 512 ? 0x8000 - 0x6 : 0
+  prg_rom_15 : ORIGIN = __prg_rom_15_lma, LENGTH = __prg_rom_size >= 512 ? 0x8000 - 0x6 : 0
+  vectors : ORIGIN = 0x10000 - 0x6, LENGTH = 6
+
+  /* CHR-ROM LMAs. */
+  chr_rom_0   : ORIGIN = 0x01000000, LENGTH = __chr_rom_size >= 8    ? 0x2000 : 0
+  chr_rom_1   : ORIGIN = 0x01002000, LENGTH = __chr_rom_size >= 16   ? 0x2000 : 0
+}
+
+REGION_ALIAS("c_readonly", prg_rom_0)
```
mysterymath, line 63: A silent invariant that we've maintained is that default C sections code must go into a fixed region. The C runtime broadly assumes that libcalls are equally accessible from any function, and libcalls will intrinsically always be placed into the base C sections. More broadly, the C runtime model basically assumes flat addresses; so the C enclave in a environment where that isn't true should generally be the maximal subset of that environment that can be made to have flat addresses.

A way of using C that doesn't invoke libcalls isn't formally specified, and we probably shouldn't start. The implications of all this would be that there is no guaranteed safe way of writing C in contexts where the libcalls might be banked out.

A far code model would help with this; the C flat addressing model can be restored if addresses contain the information necessarily to silently bank libcalls in and out.

Ostensibly this is true for RAM used by libcalls too (although I don't know whether there actually is any at present; I've tried to keep that as minimal as possible); placing the default C writeable sections in PRG-RAM can be just as dangerous. This is always an explicit action on the part of the user though, and unbanked PRG-RAM poses no obstacle for this.

Another thing we could do is carefully place libcalls in sections naming them as such; then mappers like this one could keep them fixed while allowing the rest to vary. Up to you whether it's worth it for this one; we could go back and do this later. Even with a far code model, that might still be a good idea; it would allow summarily marking libcalls near, saving banking overhead for common operations.

mos-platform/nes-gtrom/_prg-rom-banked.ld
```diff
@@ -0,0 +1,259 @@
...
+    *(.prg_rom_fixed .prg_rom_fixed.*)
+    __prg_rom_fixed_end = .;
+    __prg_rom_fixed_size = __prg_rom_fixed_end - __prg_rom_0_lma;
+  } >prg_rom_0
+
+  /* Offset each non-fixed bank section's LMA by the fixed section. */
+  /* This is accounted for in the custom output format. */
+  .prg_rom_0 : {
+      /* Non-fixed bank */
+      . = __prg_rom_0_lma + __prg_rom_fixed_size;
```
mysterymath, line 77: Does this work as an address set before the colon in the section description? That's generally more idiomatic.

mos-platform/nes-gtrom/_prg-rom-banked.ld
asiekierka: I apologize, this should have been more documented: Everything that is placed in the `prg_rom_0` region *before*  the `.prg_rom_fixed : {` section is declared (yes, in file parsing order) will become part of the faux "fixed" bank. This works correctly, as all the C sections are allocated to the `c_readonly` region; the PRG-ROM swapping tests rely on this behaviour.

The invariant is correct to maintain, I just forgot to document it.

mos-platform/nes-gtrom/_prg-rom-banked.ld
asiekierka: Using `AT()`? No, it appears that it does not work - the functions get placed at `__prg_rom_%d_lma`, ignoring the fixed bank offset.

mysterymath: Hm? No, just as `.prg_rom_0 __prg_rom_0_lma + __prg_rom_fixed_size : { }`

mos-platform/nes-gtrom/_prg-rom-banked.ld
asiekierka: Added a comment which provides the necessary clarification.

mos-platform/nes-gtrom/_prg-rom-banked.ld
asiekierka: Yes, you're right. That works fine. I have also learned that I have been writing too many dots in my link scripts :-)

mos-platform/nes-gtrom/_prg-rom-banked.ld
```diff
@@ -0,0 +1,252 @@
...
+  } >prg_rom_0
+
+  /* Offset each non-fixed bank section's LMA by the fixed section. */
+  /* This is accounted for in the custom output format. */
+  .prg_rom_0 __prg_rom_0_lma + __prg_rom_fixed_size : {
+      /* Non-fixed bank */
+      *(.prg_rom_0 .prg_rom_0.*)
+      
+      /* Non-fixed DPCM */
+      . = ABSOLUTE(.) < 0xc000 ? 0xc000 : ALIGN(64);
```
mysterymath, line 89: Ah, similarly, there was recently cleanup work done to `dpcm.ld` to break out `.dpcm` into its own section. Right now, any filler inserted by the set to `.` will be accounted for in the linker map to `.prg_rom_0`, obscuring how much was actually used. Similarly, such sections can be defined as `.dpcm_k (ABSOLUTE(.) < 0xc000 ? 0xc000 : .) : ALIGN(64) {}`

mysterymath: Sorry, this should be `.dpcm_k (ABSOLUTE(.) < 0xc000 ? 0xc000 : ALIGN(64)) {}`; it doesn't automatically align an explicit address.

mos-platform/nes-gtrom/_prg-rom-banked.ld
mysterymath: Sorry for diving right into a whole spiel; I should have probably asked a question here. 😅

mos-platform/nes-gtrom/_prg-rom-banked.ld
asiekierka: It turns out they needed a bit more fixing than that, but hey, there's tests now!

mysterymath (approved):
Alright, LGTM!

Issue #162, opened 2023-08-30
ATASCII character sets


asiekierka:
The Atari 8-bit machines' ATASCII is partially compatible with ASCII, but its drawing characters and lack of certain symbols could be better supported. Mappings to Unicode exist online.

There are four variants:

* graphics/international,
* video/interchange (including/excluding control codes).

Issue #160, opened 2023-08-30
Commander X16 PETSCII character set


mysterymath:
The Commander X16 supports PETSCII, but there is no official Unicode mapping for this target. We should go through and compare it to the existing Commodore targets to produce an accurate Unicode mapping. Hopefully this is identical to that of the Commodore 64; if not, the mapping should be given as a table on the wiki. This should also include the ISO character set; llvm-mos's execution character set is UTF-8, which is a multibyte encoding, so a user-defined literal for this is still very useful, even though the mapping is trivial.

asiekierka:
The Commander X16's changes relative to the Commodore 64 are documented here:

https://github.com/commanderx16/x16-rom/blob/master/charset/petscii.s

In addition, an additional character set based on ISO-8859-15 is supported, documented here:

https://github.com/commanderx16/x16-rom/blob/master/charset/iso-8859-15.s

This is referred to as "ISO mode" in the changelogs.

asiekierka:
The Unicode mapping for ISO-8859-15 in a familiar format is available here, but it doesn't include the CX16 reverse video additions:

https://unicode.org/Public/MAPPINGS/ISO8859/8859-15.TXT

PR #201, opened 2023-09-15
atari8, cx16: Add charset support.

asiekierka:
Fixes https://github.com/llvm-mos/llvm-mos-sdk/issues/162 ; fixes https://github.com/llvm-mos/llvm-mos-sdk/issues/160 ; also consistently handles non-prefixed and lowercase-`u`-prefixed strings in atari8 and c64 targets, as the compiler-given error can be somewhat unintuitive otherwise.

### DIFF
--- a/mos-platform/atari8-common/CMakeLists.txt
+++ b/mos-platform/atari8-common/CMakeLists.txt
@@ -12,6 +12,7 @@ install(FILES
   _pia.h
   _pokey.h
   atari.h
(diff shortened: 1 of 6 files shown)


mysterymath (approved):
Awesome, it's a lot of work, right? Or did you write some fancy scripts? :)

asiekierka:
I wrote a very non-fancy script, of course ;-)

PR #206, opened 2023-09-17
Allow for more kinds of VRAM_BUF operations

cogwheel:
This restructures the multi_vram_* headers to use a byte for the opcode rather than overloading with the upper byte of the VRAM address. This makes it possible to add many more types of operations to the VRAM buffer in the future (some suggestions are shown in the comments).

For a 3-byte `multi_vram_buffer` call, this shaved 1 cycle of the initial call, and 14 cycles off the call to `flush_vram_update2` during NMI.


cogwheel:
Superseded by #208

PR #207, opened 2023-09-17
Added VERA FX registers to cx16.h

ToboterXP:
No comments on the registers, because they're quite complicated


mysterymath (approved):
Thanks for the contribution! This LGTM, but I'll want to wait on the cc65 version to merge; they may have comments, and our implementation should match theirs as far as possible.

mysterymath:
TODO: Watch https://github.com/cc65/cc65/pull/2196

mysterymath:
Apparently they take a lot longer than we do on PRs, so I'm happy to be the first to take this in. We can adjust things later if need be. LGTM!

PR #212, opened 2023-09-20
Improve NMI performance for multi_vram_buffer ops

cogwheel:
For a single 3-byte horizontal operation this saved 14 cycles from `flush_vram_update2`

With two `multi_vram_buffer_vert` operations totaling 36 tiles and 10 `one_vram_buffer` this saved 44 cycles. This looks like it works out to ~12 cycles per multi and 2 cycles per single.

### DIFF
--- a/mos-platform/nes/nesdoug/nesdoug.s
+++ b/mos-platform/nes/nesdoug/nesdoug.s
@@ -26,32 +26,40 @@ __post_vram_update:
 .section .text.multi_vram_buffer_horz,"ax",@progbits
 .globl multi_vram_buffer_horz
 multi_vram_buffer_horz:
-
-	sta __rc5 ; loop count
+	;     A - len
+	;     X - <ppu_address
+	; __rc2 - <data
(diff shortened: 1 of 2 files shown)


mysterymath (commented):
Wouldn't this break the API of neslib's set_vram_buffer? Tweaks to internal only data structures make sense to improve performance here and there, but it looks like the nature of the buffer itself is part of the API. Correct me if I'm wrong though; it's been a while since I've looked at this.

cogwheel:
Hmm... I was under the impression that the function just changes the address of the vram buffer and that the existing APIs would still be used (though I should've updated that comment).

But yes, if a project has custom code to insert operations into the buffer, then this will be a breaking change.

I'll see if I can keep some of the performance gains without the format change.

cogwheel:
I guess most of the gains were elsewhere; going back to the old format only cost 2 cycles for the multi_vram_buffer ops.

mysterymath (approved):
LGTM!

PR #217, opened 2023-09-22
Split neslib and nesdoug components so they are linked conditionally

### LINKED ISSUE #214: Split out VRAM buffer parts of neslib/nesdoug
Scheduling updates to VRAM is a problem with many possible approaches with different tradeoffs. The approach in neslib/nesdoug isn't well separated from the rest of the library, so making a different choice here tends to require throwing away the rest of the library. Instead, this should be broken out such that inclusion of the NMI update and data structures depends on the functionality actually being used.

cogwheel:
Fixes #214


mysterymath:
> Addresses #214

Nitpick, if you say the "Fixes" instead of "Addresses", Github will automatically close the issue on merge; very nice to have.
See https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword for details.

cogwheel:
> EDIT: That list is quite long, you just *happened* to choose a word that isn't on it. I blame the list, get the thesaurus!

This is very on-brand for me. ><

mos-platform/nes/neslib/ntsc.s
```diff
@@ -3,9 +3,15 @@
 .include "ntsc.inc"
 .include "neslib.inc"
 
-.section .init.275,"axR",@progbits
+.section .init.330,"axR",@progbits
```
mysterymath, line 6: We should maintain the property that the neslib initializers are done before `.init.300`; that's when user-defined module level constructors fire, and broadly anything that is accessible in C should work properly from those pre-main constructors, including the NTSC mode.

This is true here and throughout this file; anything that previously occurred before `.init.300` should stay before `.init.300`.

cogwheel: looks like i forgot to include `mos-platform/common` when I was searching for existing init ranges.

mysterymath (approved):

mysterymath:
Unfortunately, I ended up having to revert this; see the above commit message for details.

Issue #216, opened 2023-09-21
Add sections for the ram memory area when c-in-prg-ram is used


cogwheel:
On NES mapper targets, when `c-in-prg-ram` is used, there are no sections that map into the RAM memory area. This makes (2 KiB - 2 pages) worth of RAM effectively unavailable for use without adding custom linker scripts.

Maybe there could sections like `.low_ram.(bss|data|noinit).*` that always map to the `ram` region? These could even be used by things like the neslib buffers and the software stack to leave more "user" ram available (and marginally simplify PRG-RAM banking)

mysterymath:
By analogy these should just be called `.ram`; this was pretty much just a straight oversight. But yes, every other region has dedicated sections; RAM should too.

mysterymath:
Fixed by #218.

PR #218, opened 2023-09-22
Add .ram section for NES system RAM

cogwheel:
Addresses #216


mos-platform/nes/nes.ld
```diff
@@ -42,5 +42,7 @@ SECTIONS {
   INCLUDE bss.ld
   INCLUDE noinit.ld
 
+  .ram (NOLOAD) : { *(.ram .ram.*) } > ram
```
mysterymath, line 45: Nice on the NOLOAD! LGTM.

mysterymath (approved):

PR #222, opened 2023-09-25
Improve allocation of VRAM/PAL/OAM buffers

### LINKED ISSUE #215: Provide linker-configured PAL_BUF for neslib/nesdoug
Currently, the palette buffer used by neslib/nesdoug is hard-coded to the low portion of the hardware stack. This should be made a weak symbol instead to allow for the user to place it elsewhere. Stack is particularly valuable on the NES, since it enables techniques like pop-slide.

### LINKED ISSUE #219: Suggestion: Keep some things in RAM when c-in-prg-ram is used
In order to ~~a)~~ maximize the RAM available to ordinary front-end code ~~and b) facilitate PRG-RAM banking~~, I propose keeping the following data in the `.ram` section when `c-in-prg-ram` is used:

  - [ ] software stack (#223)
  - [ ] oam_buf
  - [ ] pal_buf
  - [ ] vram_buf
  - maybe other system-level data I'm unaware of?
  
  This would go along with making them weak symbols to allow customization.

cogwheel:
Fixes #215:
  - Use weak symbols
  - Move PAL_BUF out of hardware stack

### DIFF
--- a/mos-platform/nes/nesdoug/vram_buffer.c
+++ b/mos-platform/nes/nesdoug/vram_buffer.c
@@ -3,11 +3,12 @@
 
 __attribute__((section(".zp.vram_index"))) char VRAM_INDEX;
 
-extern char VRAM_BUF[];
+extern char VRAM_BUF[128];
+
+__attribute__((leaf)) void __post_vram_update();
 
 void set_vram_update(const void *buf);
 void set_vram_buffer(void) {
-  VRAM_BUF[0] = 0xff;
-  VRAM_INDEX = 0;
   set_vram_update(VRAM_BUF);
+  __post_vram_update();
 }
--- a/mos-platform/nes/nesdoug/vram_buffer.s
+++ b/mos-platform/nes/nesdoug/vram_buffer.s
@@ -4,9 +4,8 @@
 .zeropage VRAM_INDEX
 
 .section .noinit.vram_buf,"aw",@nobits
-.globl VRAM_BUF
+.weak VRAM_BUF
 .balign 128
-.weak
 VRAM_BUF:
 	.zero 128
 
--- a/mos-platform/nes/nesdoug/vram_buffer.s
+++ b/mos-platform/nes/nesdoug/vram_buffer.s
@@ -6,8 +6,9 @@
 .section .noinit.vram_buf,"aw",@nobits
 .globl VRAM_BUF
 .balign 128
+.weak
 VRAM_BUF:
-  .zero 128
+	.zero 128
 
 .text
 .global __post_vram_update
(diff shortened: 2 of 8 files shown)


cogwheel:
Hmm this works with 26_Full_Game but not 21_Sprite_Zero.

cogwheel:
Regarding https://github.com/llvm-mos/llvm-mos-sdk/pull/222/commits/b5d3522079bd51c39fda0a3b0b3c1a6358ff7a41 I'm curious about the reasoning for `__post_vram_update`. Is there ever a time the buffer shouldn't be cleared and the counter reset after writing it to vram?

mos-platform/nes/nesdoug/vram_buffer.c
```diff
@@ -2,12 +2,14 @@
 // version 1.2, 1/1/2022
 
 __attribute__((section(".zp.vram_index"))) char VRAM_INDEX;
+__attribute__((weak, aligned(128), section(".noinit.vram_buf"))) char VRAM_BUF[128];
 
-extern char VRAM_BUF[];
-
-void set_vram_update(const void *buf);
-void set_vram_buffer(void) {
+void __post_vram_update() {
```
mysterymath, line 7: This must be callable from and interrupt context, and right now C isn't unless it's `__attribute__((interrupt))` or `__attribute__((interrupt_norecurse))`. Without the AXY calling convention, these will blow up the sizes of the routine, so for now, this needs to stay in assembly.

mysterymath:
> Regarding [b5d3522](https://github.com/llvm-mos/llvm-mos-sdk/commit/b5d3522079bd51c39fda0a3b0b3c1a6358ff7a41) I'm curious about the reasoning for `__post_vram_update`. Is there ever a time the buffer shouldn't be cleared and the counter reset after writing it to vram?

IIRC it was just that VRAM_BUF is a nesdoug addition to neslib, and it shouldn't trigger unless nesdoug is used.

mos-platform/nes/nesdoug/vram_buffer.c
cogwheel: Hmm... I was following the example of `__early_init`/`__late_init`. For NMI, I was also under the impression a void(void) function that only does basic memory ops on global variables wouldn't need anything besides A, X, and Y, which are preserved by the normal NMI preamble.

But yeah, having a c function that can't be used as a normal C function is probably less good than just duplicating the code.

mysterymath: > Hmm... I was following the example of `__early_init`/`__late_init`. For NMI, I was also under the impression a void(void) function that only does basic memory ops on global variables wouldn't need anything besides A, X, and Y, which are preserved by the normal NMI preamble.

This is likely to be true, but nothing guarantees this, and code generator flak may sneak in imaginary register references. It's best not to depend on the specific output of the code generator for given C; one should instead rely upon ABI guarantees. That's why I'm hoping to eventually make "only touches AXY" an ABI guarantee: so we can rely on it.

cogwheel: Hmm... the commit where I moved it into C said it was to fix something. I need to remember what that was 🤔

cogwheel: Oh ... `21_sprite_zero` calls `set_vram_update` but doesn't use any of the functions that are defined in `vram_buffer.s`. It only uses `metatile.c/s`. 

So I can either have `set_vram_buffer` call the assembly version of `__post_vram_update` or i can separate `multi/one-vram-buffer` functions from the buffer definition.

mos-platform/nes/nesdoug/vram_buffer.s
```diff
@@ -6,8 +6,9 @@
 .section .noinit.vram_buf,"aw",@nobits
 .globl VRAM_BUF
 .balign 128
+.weak
```
mysterymath, line 9: This doesn't do anything; the syntax is `.weak <names>`. So, this should be `.weak VRAM_BUF`.

cogwheel: Oh hm... apparently `.weak` does the job of `.globl` too. Fixed.

mysterymath (approved):

PR #231, opened 2023-10-24
[PET, VIC20] Add target, add cc65-compatible header files.

asiekierka:


mos-platform/vic20/link-5k.ld
```diff
@@ -0,0 +1,23 @@
...
+ * BASIC memory regions are freely overwritten, but any KERNAL regions are left alone.
+ *
+ * Produces a PRG file with a SYS command to start the program.
+ */
+
+__basic_zp_start = 0x0000;
+__basic_zp_end = 0x0090;
+
+MEMORY {
+    ram (rw) : ORIGIN = 0x1001, LENGTH = 0x0DFF
```
mysterymath, line 12: Could we have this symbol-controlled instead of having separate linker scripts? If there's only 3 cases the logic should be fairly straightforward.

mos-platform/vic20/link-5k.ld
```diff
@@ -0,0 +1,23 @@
...
+}
+
+INCLUDE commodore.ld
+
+INPUT(init-stack-memtop.o)
+
+OUTPUT_FORMAT {
+    /* Tells the VIC-20 LOAD command where to place the file's contents. */
+    SHORT(0x1001)
```
mysterymath, line 21: SHORT(ORIGIN(ram))

mos-platform/vic20/link-5k.ld
asiekierka: Should be fixed now.

mos-platform/vic20/link-5k.ld
asiekierka: Should be fixed now for all C= targets.

mysterymath (approved):

PR #233, opened 2023-10-26
[C128, PET, VIC20] Add C128 target, minor PET/VIC20 fixes.

asiekierka:
Mostly modeled off the cc65 target.


mos-platform/c128/link.ld
```diff
@@ -0,0 +1,25 @@
+/* C128 PRG file linker script.
+ *
+ * BASIC memory regions are freely overwritten, but any KERNAL regions are left alone.
```
mlund, line 3: I may have missed it, but how is the basic rom disabled?

asiekierka: `init-mmu.s` configures bank 0 to have the KERNAL enabled while disabling BASIC HI/LO. See http://www.oxyron.de/html/registers_mmu.html

The best option would be to disable the KERNAL ROM and do far jumps to it, which would unlock ~15.75KB of extra contiguous memory for C code.

mos-platform/c128/init-mmu.S
```diff
@@ -0,0 +1,15 @@
+.section .bss,"aw",@nobits
```
mysterymath, line 1: nit: Idiomatically, `.bss`.

mos-platform/c128/init-mmu.S
```diff
@@ -0,0 +1,15 @@
...
+__mmusave:
+  .fill 1
+
+; MMU initialization needs to complete before any CRT initialization
+.section .init.010,"axR",@progbits
+        lda $FF00
```
mysterymath, line 7: Instead, we can include `c128.inc` and do this symbolically; that should make the logic here very obvious

mysterymath (approved):

mos-platform/c128/init-mmu.S
asiekierka: That required actually fixing all the include files, but is now fixed.

PR #237, opened 2023-11-07
Added initial Atari 2600 support (4K and 3E mappers)

sehugg:
This is an initial attempt at an Atari 2600 target based on CC65 VCSLib: https://8bitworkshop.com/docs/posts/2023/vcslib-c-programming-for-atari-2600.html

It only supports 4K (non-banked) and 3E (2K fixed, 2K banked ROM+RAM) mappers for now.

One wrinkle with 3E is that the "write port" for RAM is $400 bytes above the "read port". So to use this area like a normal data segment, you'd have to declare a pair of variables, one read-only and one write-only and the compiler would have to make sure they are $400 bytes apart.

The 32 bytes of virtual registers is kind of a lot for the VCS but I assume the clobber attribute will let asm developers safely use this space. I'm guessing there won't be a soft stack.

The shared ROM area will be a scarce resource, especially if it is 2 KB or less. For banked mappers, it would be good to put all initialization code into rom0.

Display kernels will still probably be written in asm, except for very simple ones.

Since 2600 carts have no header, auto-detection of 3E by Stella et al is iffy. When most of the banks are empty, it triggers the wrong mapper: https://github.com/stella-emu/stella/blob/d6224a8a6e30b4b323cde86ade2f05f75dcdfbec/src/emucore/CartDetector.cxx#L294


asiekierka:
Just to verify, what is the license on VCSLib-derived code?

mos-platform/atari2600-3e/link.ld
```diff
@@ -0,0 +1,86 @@
...
+MEMORY {
+  rom0 (rx)     : ORIGIN = 0x1000, LENGTH = 0x800
+  rom1 (rx)     : ORIGIN = 0x3000, LENGTH = 0x800
+  rom2 (rx)     : ORIGIN = 0x5000, LENGTH = __cart_rom_size >= 8 ? 0x800 : 0
+  rom3 (rx)     : ORIGIN = 0x7000, LENGTH = __cart_rom_size >= 10 ? 0x800 : 0
+  rom4 (rx)     : ORIGIN = 0x9000, LENGTH = __cart_rom_size >= 12 ? 0x800 : 0
+  rom5 (rx)     : ORIGIN = 0xB000, LENGTH = __cart_rom_size >= 14 ? 0x800 : 0
+  rom6 (rx)     : ORIGIN = 0xD000, LENGTH = __cart_rom_size >= 16 ? 0x800 : 0
+  rom7 (rx)     : ORIGIN = 0xF000, LENGTH = __cart_rom_size >= 18 ? 0x800 : 0
+  rom8 (rx)     : ORIGIN = 0x1000, LENGTH = __cart_rom_size >= 20 ? 0x800 : 0
```
asiekierka, line 28: I don't think this will work right. ORIGIN addresses should be unique per memory section, if they are to point to different banks. The traditional scheme for LLVM-MOS mappers is to use the bottom 16 bits to match the VMA (virtual memory address) of the 6502, and the top 16 bits to designate the bank used.

sehugg:
> Just to verify, what is the license on VCSLib-derived code?

Much of it is from CC65, I see that other CC65-derived files have an Apache notice so I can add it if that's the right approach. I'm happy to license my original code under Apache 2.

mos-platform/atari2600-3e/link.ld
sehugg: I was wondering about that. Can I just make the top byte equal to the bank index, for example:

rom0 = 0x00003000
rom1 = 0x01003000
...etc

And for the fixed bank something like:

perm = 0xff00f800

mos-platform/atari2600-3e/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+platform(atari2600-3e COMPLETE PARENT atari2600)
```
mysterymath, line 1: I'll do a more thorough review in a bit, but I wanted to get this comment out early to save time. It's generally not desirable to have one complete target as a parent of another; targets should be open for extension, but nothing in a target should override or replace anything in its parent. Accordingly, all of the contents of a parent's e.g. libc are silently merged into a derived targets library of the same name. This is generally not what you want for situations like this.

Instead, both the "basic" `atari2600` and `atari2600-3e` targets should be made to derive from an incomplete `atari2600-common` target that has the functionality that is truly common between them.

I'm also not very familiar with the 2600, so I'll leave this to you: we should only call one of these targets `atari2600` if it's overwhelmingly the configuration that folks in the community think of when developing for it. (I'm guessing it's the most common commercial system cartridge size?)

mysterymath:
Amazing to see a 2600 port, with a C demo! I remember 7 or 8 years ago having a conversation with a 2600 scener at a silicon valley VR meetup that I was working on a 6502 C compiler, and him mentioning that it was doubtful to be possible to get even simple things working on that platform with C. This was before llvm-mos by quite a bit, and codegen still isn't 100% what we'd need for this to be completely practical, but it's still neat to see something like this!

mysterymath:
> The 32 bytes of virtual registers is kind of a lot for the VCS but I assume the clobber attribute will let asm developers safely use this space. I'm guessing there won't be a soft stack.

Yep, the virtual registers can be used arbitrarily in C, so long as the calling convention is followed. That is, C code assumes anything it calls preserves RC2 to RC19, assembly can rely on the this, but must ensure it for its callers. Similarly, assembly can arbitrarily clobber any other imaginary register, but it must assume any C routines it calls can do the same.

mos-platform/atari2600-3e/CMakeLists.txt
sehugg: I thought I could get away with a shared crt0, but I think it's a good idea to make a common parent. The 4K mapper is closest to NES NROM in that it's basically no mapper. But although it's the most straightforward, it's not the most useful for new development, so I think explicitly calling it atari2600-4k is good.

I was having some intermittent problems with the init code going away unless I added an optimization flag (?) but that seems fixed, maybe it was a CMake thing. Also not quite sure on weak functions, they were sometimes getting zeroed out when I tried to invoke them.

Since 2600 carts have no header, emulators use a bunch of heuristics to determine the mapper, and it doesn't always work right. 4K is fine, but current emulators will probably have to be told the 3E example ROM is 3E. Maybe there can be a "autodetection helper library" to smooth things out.

mos-platform/atari2600-3e/clang.cfg
```diff
@@ -0,0 +1 @@
+-D__ATARI2600_MAPPER_3E__
```
mysterymath, line 1: Add a newline at end of file

mos-platform/atari2600-3e/link.ld
```diff
@@ -0,0 +1,129 @@
...
+
+/*
+   High word of origin address:
+   0x00 0x00    = fixed ROM
+   0x01 [bank#] = bank-switched ROM
+   0x02 [bank#] = bank-switched RAM
+*/
+
+MEMORY {
+  rom0 (rx)     : ORIGIN = 0x01001000, LENGTH = 0x800
```
mysterymath, line 27: There's not much use setting these attributes, since these only control automatic placement of sections that aren't mentioned in the linker script. Due to the arcanity of those rules, we tend to require that all sections be explicitly mentioned in `SECTIONS`. In any case, only the first memory region with a given attribute set could possibly have any effect; there's no automatic spill-over once that one fills up; instead the link fails.

mos-platform/atari2600-3e/mapper.h
```diff
@@ -0,0 +1,16 @@
+/* TigerVision (3E) mapper include files */
```
mysterymath, line 1: Add include guards

mos-platform/atari2600-3e/mapper_3e.c
```diff
@@ -0,0 +1,73 @@
...
+// Licensed under the Apache License, Version 2.0 with LLVM Exceptions.
+// See https://github.com/llvm-mos/llvm-mos-sdk/blob/main/LICENSE for license
+// information.
+
+#include <peekpoke.h>
+#include <mapper.h>
+
+__attribute__((section(".zp.bss"))) bank_index_t __current_bank;
+
+__attribute__((weak)) void set_current_bank(bank_index_t bank_id) {
```
mysterymath, line 11: Does it actually make sense to have these weak? Can the functionality in this file be meaningfully partially overriden?
I tended to go a bit weak-happy in the early days of the SDK, but library files like this will only be brought in to the link if an undefined symbol reference is fulfilled by it, so all weak buys is allowing you to replace part, but not all, of this library.

mos-platform/atari2600-3e/mapper_3e.c
```diff
@@ -0,0 +1,73 @@
...
+  if (bank_select && !(bank_id & 0x80)) {
+    bank_select(bank_id);
+  } else if (ram_select) {
+    ram_select(bank_id & 0x7f);
+  }
+}
+
+__attribute__((always_inline, leaf)) bank_index_t get_current_bank(void) {
```
mysterymath, line 19: Leaf does nothing on a function definition; it's only use is on declarations of functions external to a LTO (link-time optimization) unit.

Also always_inline is generally not a good idea unless there's a compelling reason; the inliner generally does a good job of deciding whether to inline cases like these, and always_inline can interfere with e.g. taking the address of the function.

mos-platform/atari2600-3e/mapper_3e.c
```diff
@@ -0,0 +1,73 @@
...
+  POKE(0x3f, bank_id);
+  __current_bank = bank_id;
+}
+
+__attribute__((always_inline, leaf)) void ram_select(bank_index_t bank_id) {
+  POKE(0x3e, bank_id);
+  __current_bank = bank_id | 0x80;
+}
+
+__attribute__((leaf)) void xram_write(int offset, unsigned char value) {
```
mysterymath, line 33: Note: char is unsigned by default in llvm-mos. No action required; you can continue to be explicit.

mos-platform/atari2600-3e/mapper_3e.c
```diff
@@ -0,0 +1,73 @@
...
+// See https://github.com/llvm-mos/llvm-mos-sdk/blob/main/LICENSE for license
+// information.
+
+#include <peekpoke.h>
+#include <mapper.h>
+
+__attribute__((section(".zp.bss"))) bank_index_t __current_bank;
+
+__attribute__((weak)) void set_current_bank(bank_index_t bank_id) {
+  if (bank_select && !(bank_id & 0x80)) {
```
mysterymath, line 12: This is checking whether the address of bank_select is null, but it's defined in the same file, so it could never be. Is something else intended here?

mos-platform/atari2600-4k/clang.cfg
```diff
@@ -0,0 +1 @@
+-D__ATARI2600_MAPPER_4K__
```
mysterymath, line 1: nit: newline at end of file

mos-platform/atari2600-common/mapper_rom_multi.h
```diff
@@ -0,0 +1,29 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// ROM bank index type.
+typedef unsigned char rom_bank_t;
+typedef unsigned char banked_area_t;
+
+// Switch in a ROM bank into a given area.
+__attribute__((weak, leaf)) void bank_select(
```
mysterymath, line 21: leaf has no effect on declarations that are later fulfilled by functions in the same LTO unit (typically, all those defined in C).

mos-platform/atari2600-common/mapper_xram_single.h
```diff
@@ -0,0 +1,48 @@
...
+// RAM bank index type.
+typedef unsigned char ram_bank_t;
+
+// Macro to declare a variable in XRAM.
+// - index = bank index
+// - declaration = variable declaration
+// This will declare two variables, one read-only and one write-only.
+// For example, DECLARE_XRAM_VARIABLE(0, int my_var)
+// creates two variables: my_var_read and my_var_write.
+#define DECLARE_XRAM_VARIABLE(index, declaration) \
```
mysterymath, line 25: Very very slick!

mos-platform/atari2600-common/vcslib.h
```diff
@@ -0,0 +1,84 @@
...
+
+// Function to set horizontal position of a game object.
+
+typedef enum { PLAYER_0=0, PLAYER_1, MISSILE_0, MISSILE_1, BALL } TIAObject;
+
+// A = X coordinate
+// X = object index
+__attribute__((leaf)) void _set_horiz_pos(byte xpos, byte objindex);
+// swap order of call
+#define set_horiz_pos(objindex,xpos) _set_horiz_pos(xpos,objindex);
```
mysterymath, line 37: Instead, just make this a regular function. The compiler should almost always inline it, but it allows function pointers to work.

mos-platform/atari2600-common/vcslib.h
```diff
@@ -0,0 +1,84 @@
...
+
+// Waits for next scanline start
+#define do_wsync() \
+        asm("sta $42 ; WSYNC");
+
+// Applies horizontal motion to sprite(s) after set_horiz_pos()
+#define apply_hmove() \
+        asm("sta $42 ; WSYNC"); \
```
mysterymath, line 45: The compiler can move arbitrary code between these; if that's ok, NAR, otherwise these should be in the same ASM statement, separated by '\n'.

mos-platform/atari2600-common/vcslib.h
```diff
@@ -0,0 +1,84 @@
...
+
+// Macros
+
+#define P0	0
+#define P1	1
+#define M0	2
+#define M1	3
+#define BALL	4
+
+#define SW_RESET()	((RIOT.swchb & RESET_MASK) == 0)
```
mysterymath, line 56: Any of the macros after this point you desire can be made regular functions, and the compiler will again almost always inline them away or evaluate them at compile time.

mos-platform/atari2600-3e/mapper_3e.c
mysterymath: Actually, I rescind the always_inline part of this; it looks like on clang at least it's not a failure if it can't actually inline; always_inline just makes it always try.

mysterymath (commented):
For any of the macro->function comments I left; it just occurred to me that you may be trying to keep a common API between this and other versions of vcslib. If that's the case, no worries; consistency trumps all.

mos-platform/atari2600-3e/mapper_3e.c
sehugg: That was due to my misunderstanding of the weak attribute, I thought it pruned functions that are never called by non-weak functions. But it turns out the optimizer figures out that bank_id is never set >= 128 and can optimize it anyway :) Anyway I removed all the weak attributes.

mos-platform/atari2600-3e/mapper_3e.c
```diff
@@ -8,24 +8,24 @@
 
 __attribute__((section(".zp.bss"))) bank_index_t __current_bank;
 
-__attribute__((weak)) void set_current_bank(bank_index_t bank_id) {
-  if (bank_select && !(bank_id & 0x80)) {
-    bank_select(bank_id);
-  } else if (ram_select) {
+__attribute__((leaf)) void set_current_bank(bank_index_t bank_id) {
```
mysterymath, line 11: To clarify; the basic rule is that no function definition in C should ever have `__attribute ((leaf))`. This is only useful for declarations of functions written assembly files, or much less commonly, C files that were intentionally compiled with `-fno-lto`.

mos-platform/atari2600-common/vcslib.h
sehugg: Strangely enough, I tried inlining this function but it refused even with always_inline. It specialized the function since it was called with one of the parameters fixed, but it didn't inline either function.

mysterymath (approved):
Thanks for the PR, LGTM!

PR #241, opened 2023-11-15
Emutest + Libretro tests (Atari 2600 + NES)

sehugg:
I made two separate test libraries, one for Mesen and one for emutest which writes an 8-byte pass/fail signature into RAM. The Atari 2600 test library also shows a nice pass/fail screen if you load the ROMs on an emulator.

There's also a mode where you pass the expected frame buffer CRC into the test. This is for Atari 2600 kernels or anything else difficult to test with assertions.

I also updated the Github Actions file -- updated to Ubuntu 22.04, installs Go, and it fetches the Stella2014 core via rpm. I forked the emutest project to make changes, I'll contribute them back to the trunk but right now the script fetches my fork.

I've got another branch with the existing NES tests running on the Nestopia core, but not all are passing.


.github/workflows/package.yml
```diff
@@ -12,13 +12,17 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [windows-2019, macos-latest, ubuntu-20.04]
```
mysterymath, line 15: We can't do this; we need to build against the oldest version of a dynamically linked userspace possible, since glibc guarantees only backwards compatibility, not forwards compatibility. Applications linked against old Linux can run on new Linux, but applications linked against new Linux cannot run on old Linux.

sehugg: Gotcha. I think I can make everything work on 20.04.

sehugg:
I finally got a build where all the Mesen tests pass, some were occasionally SIGSEGV-ing. Have you had this problem? Maybe I'm getting different test runner machines.

mysterymath:
I've definitely seen flakes in the test runners; if it's only occasional I
wouldn't worry about it.

sehugg:
I've got all the NES tests passing via Libretro cores now, except for two in nes-gtrom which still use Mesen. The FCEUMM core passed more tests than Nestopia, but they're both available.

I also updated the GIthub workflow to build test tools in a separate job to make easier to add more emulators. Not totally sure about the caching strategy, but for now the cache key depends on the workflow file digest.

.github/workflows/package.yml
```diff
@@ -8,8 +13,69 @@ on:
...
+          go install
+          mkdir -p ${{ env.EMUTEST_DIR }}
+          GOPATH=$(go env GOPATH)
+          cp $GOPATH/bin/emutest ${{ env.EMUTEST_DIR }}
+      - name: Fetch Mesen
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: |
+          mkdir -p ${{ env.MESEN_DIR }}
+          cd ${{ env.MESEN_DIR }}
+          #curl -Lo Mesen.exe https://github.com/NovaSquirrel/Mesen-X/releases/download/1.0.2/Mesen.exe
```
mysterymath, line 44: Add a comment about why this is commented out

.github/workflows/package.yml
```diff
@@ -8,8 +13,69 @@ on:
...
+          chmod +x mesen
+      - name: Build Stella2014 core
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: |
+          git clone https://github.com/libretro/stella2014-libretro
+          cd stella2014-libretro
+          make --silent -j 4
+          mkdir -p ${{ env.LIBRETRO_CORES_DIR }}
+          mv stella2014_libretro.so ${{ env.LIBRETRO_CORES_DIR }}
+      - name: Build Nestopia core
```
mysterymath, line 59: Does the mesen libretro core not work? If so, do we have any insight into why not? It's not exactly desirable to require three separate NES emulators to test the SDK; it would be more maintainable to use just native Mesen if we can't avoid it.

test/test.cmake
```diff
@@ -1,14 +1,28 @@
...
 
 function(add_mesen_test name)
   add_executable(${name} ${name}.c)
-  target_link_libraries(${name} test-lib)
+  target_link_libraries(${name} test-lib-mesen)
   add_test(NAME test-${name} COMMAND ${MESEN_COMMAND} --testrunner
            $<TARGET_FILE:${name}> ${CMAKE_CURRENT_SOURCE_DIR}/../mesen.lua)
 endfunction()
 
+function(add_emutest_test name binext source_dir libretro_core)
```
mysterymath, line 16: Before, the test suite could (for the most part) only be run if mesen was available, so there was little point in trying to without it installed. Now that the dependency list is growing, it should be possible to partially run the test suite with only some of the dependencies installed. This makes it possible to test and fix CI breakages locally without needing to install the whole CI world (which will become increasingly intractible).

.github/workflows/package.yml
sehugg: I did get the Mesen-X Libretro core working, and was surprised to see it failed on exactly the same two tests as FCEUMM, chr-rom and chr-swap-split. I think maybe the problem is that the GTROM 111 mapper does not support CHR ROM. @asiekierka do you have any more info on this?

Assuming we figure this out, all the tests will pass with both Mesen-X and FCEUMM cores, and we can lose native Mesen and Nestopia. I think it's good to have two NES emulators in case a test author wants to compare results.

test/test.cmake
sehugg: Yep, CMake finds all Libretro cores and will fail tests where you don't have them installed. You can run e.g. "ninja test-nes-nrom" for example to run a single test project. I could add a short section to the README with links to dependencies and instructions.

.github/workflows/package.yml
asiekierka: There is no "mapper which doesn't support CHR-ROM". Nothing prevents you from wiring ROM in place of RAM in this regard; it's not like PRG-ROM and PRG-RAM. If an emulator doesn't emulate that combination, it's the fault of the emulator.

mysterymath (approved):
Alright, I finally found time to sit down and give this a read. Thanks for putting this together, great PR! This is tremendously good for the long term health of the project, since we've not really had much of a strategy for dealing with SDK regressions outside the NES, but now we do, and the frame CRC solution broadens what that can cover.

test/emutest.lua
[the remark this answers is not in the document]
sehugg: Glad you find it useful; emutest was a nice find! I removed Nestopia and replaced it with the Mesen-X core, all the tests are passing on my end. The Libretro cores were surprisingly well-behaved, hopefully this holds for future systems. I also added a few notes to the docs to entice future devs to write tests ;)

Issue #242, opened 2023-11-15
cx16 kernal name confusion: cx16_k_savehl vs cbm_k_savehl


bcampbell:
`cx16_k_savehl` is declared in `cx16.h` and implemented in `mos-platform/cx16/cx16_k_savehl.s`.

`cx16_k_savehl` is exported, but the actual label used on the routine is `cbm_k_savehl`. So when you try and link, it can't find it.

Also, the cx16 kernal docs use the name BSAVE instead (same entry address as SAVEHL: $FEBA). The cx16 [kernal docs](https://github.com/X16Community/x16-docs/blob/master/X16%20Reference%20-%2004%20-%20KERNAL.md#kernal-api-functions) claim BSAVE is a cx16-specific routine, but now I'm wondering if other commodore kernals (128/+4/etc) had it as SAVEHL?

mysterymath:
@XarkLabs

XarkLabs:
I will look into this.  This HL is like BSAVE, but without a two byte header IIRC...

XarkLabs:
This cx16 call was previously undocumented and I was using internal name "savehl", now it is properly documented as BSAVE, so #243 renames llvm-mos-sdk cx16 function and vector to use documented name.

bcampbell:
Ahh, thanks for that! Both for the fix and for casting some light on the history of it!
Much appreciated.

XarkLabs:
My pleasure. Thanks for mentioning it (and also pointing out my "typo" in the older version).

PR #243, opened 2023-11-16
Rename cx16 kernel call savehl to bsave

XarkLabs:
This call was not documented previously, and internal vector was called savehl (save headerless).  Now it is documented as BSAVE so this PR renames kernel vector and function to match documentation.


mysterymath:
Fixes #242

mysterymath (approved):

PR #250, opened 2023-11-26
Atari 8-bit XEGS (bank-switched) cartridge support

cwedgwood:
Support for XEGS cartridges from 32K to 512K.


cwedgwood:
@asiekierka, @mysterymath 
This is just an RFC based on something I had working a very long time ago.

asiekierka:
This is not how we do things anymore. There was a refactor which unified LMA and VMA addresses: https://github.com/llvm-mos/llvm-mos-sdk/commit/18e0edd90648f67717b26f99559dae408eaa8b32

cwedgwood:
> This is not how we do things anymore. There was a refactor which simplified LMA and VMA address handling, instead opting to use `OUTPUT_FORMAT` to concatenate banks: [80b355d](https://github.com/llvm-mos/llvm-mos-sdk/commit/80b355d777adc50674b7e7122ce568fa69bb99cc)
> 
> This approach is more linker-friendly, and as such it is nowadays preferred.
> 
> I wouldn't merge this with stdcart unless this type of bank switching is standard across the Atari 8-bit platform.

Thanks! PTAL.

mos-platform/atari8-xegs/link.ld
```diff
@@ -0,0 +1,333 @@
...
+  .bank55    : { *(.bank55    .bank55.*)    } >bank55
+  .bank56    : { *(.bank56    .bank56.*)    } >bank56
+  .bank57    : { *(.bank57    .bank57.*)    } >bank57
+  .bank58    : { *(.bank58    .bank58.*)    } >bank58
+  .bank59    : { *(.bank59    .bank59.*)    } >bank59
+  .bank60    : { *(.bank60    .bank60.*)    } >bank60
+  .bank61    : { *(.bank61    .bank61.*)    } >bank61
+  .bank62    : { *(.bank62    .bank62.*)    } >bank62
+
+ /* "SIZE=nnn"; used when testing linker scripts, makes hexdump useful */
```
mysterymath, line 251: The blessed way to do this is to generate a linker map file: -Wl,`-Map=<mapfile>`. One can also obtain size information using llvm-readelf on the .elf file produced by the link; accordingly, it seems better to get the 8 bytes back and have a more typical linker script.

mos-platform/atari8-xegs/syms.s
```diff
@@ -0,0 +1,13 @@
+; xegs ROM size, only 32, 64, 128 and 256 are valid sizes (in KiB)
+
+.weak __cart_rom_size
+; default to 256K
+__cart_rom_size = 256
+
+; explicitly select bank 0, allows for override of INIT
```
mysterymath, line 7: What about selecting bank zero allows overriding INIT?

mos-platform/atari8-xegs/syms.s
```diff
@@ -0,0 +1,13 @@
...
+.text
+	.global _cart_init
```
mysterymath, line 9: Does it make sense for a user to provide a different version of this? If so, the symbol should be weak.

mos-platform/atari8-xegs/link.ld
cwedgwood: This is a hack I was using when testing/debugging the linker script, `hexdump -C` / `hd` is useful when the images are small/sparse to eyeball what's going on.  I will remove this.

mos-platform/atari8-xegs/syms.s
mysterymath: Is there a reason not to allow the user to define a different _cart_init? That would be worth writing down here as a comment if there is one.

mos-platform/atari8-xegs/syms.s
cwedgwood: Poor wording in my part.

mos-platform/atari8-xegs/syms.s
cwedgwood: Yes.

mysterymath (approved):

mysterymath:
Thanks for the PR, LGTM!

Issue #261, opened 2023-12-08
Trying to define a smaller MMC3 ROM results in a wrong sized output


wendelscardua:
I'm making a MMC3 ROM with mostly C++, and if I compile it along a `config.s` with these definitions:

```
        .global __chr_rom_size
        .global __prg_rom_size
        .global __prg_ram_size
        .global __four_screen
        .global __mirroring

                                ; Kilobytes
        __prg_rom_size = 64
        __chr_rom_size = 8
        __prg_ram_size = 8

                                ; Flags
        __four_screen = 0
        __mirroring = 1 ; horizontal mirroring

```

I get a ROM with 81936 bytes instead of 73744 (16 header + 65536 PRG + 8129 CHR)

(for now I'm omitting the `_prg_rom_size` part and living with a mostly empty 500K+ ROM)

mysterymath:
Yeah, it looks like there's an off-by-one error in the MMC3 linker scripts. I'll add a unit test for this regression and correct the issue.

mysterymath:
This was an off by one in the MMC3 linker script; depending on the bank layout, the fixed region may use take either 2 or 3 banks. The default setup uses 3 but neglects to subtract out a swappable bank for any size other than the max. I've added regression tests for this behavior for all of the supported size boundaries.

PR #269, opened 2023-12-13
Fix off by one in MMC3 8k banking linker script

mysterymath:
This included one too many PRG-ROM sections before the 24KiB fixed
region, which caused the ROM to be an invalid size.


mysterymath:
Fixes #261

PR #276, opened 2023-12-30
Atari 8-bit MegaCart (bank-switched) cartridge support

cwedgwood:


mos-platform/atari8-mega/link.ld
```diff
@@ -0,0 +1,372 @@
...
+  On power-up the bank0 is selected, however, SIC! cartridges-up will
+  not have the lower 8 KiB enabled.  We use INIT to fix this and
+  adjust MEMTOP if needed.
+
+  https://github.com/atari800/atari800/blob/ATARI800_5_0_0/DOC/cart.txt
+  has details. */
+
+INPUT(tail0.o)
+
+/* Default cartridge size.  This isn't the largest possible size to
```
mysterymath, line 19: Instead of defaulting to a non-extremum size, adding test for successful compilation would be a more principled solution. Is there another reason to justify 128 KiB specifically as the default size?

cwedgwood: Can do. PTAL.

PR #277, opened 2023-12-30
NES GTROM; support smaller sized ROMs

cwedgwood:
By default we test and build with the largest possible size.  The linker script does have partial support for smaller sizes, this permits those to generate the correct output.

### DIFF
--- a/mos-platform/nes-gtrom/prg-rom.ld
+++ b/mos-platform/nes-gtrom/prg-rom.ld
@@ -1,8 +1,6 @@
 INCLUDE _prg-rom-banked.ld
 
-/* Default to something smaller than the maximum size to avoid masking
-   linker script offset calculation problems. */
-PROVIDE(__prg_rom_size = 128);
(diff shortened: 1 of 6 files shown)


asiekierka:
Oops! My bad, thank you.

LGTM - for extra points, do you think it would be possible to validate the expected ROM filesize in bytes at the tests level?

mos-platform/nes-gtrom/prg-rom.ld
```diff
@@ -1,25 +1,27 @@
 INCLUDE _prg-rom-banked.ld
 
-PROVIDE(__prg_rom_size = 512);
```
mysterymath, line 3: Here it still makes sense to use an extreme size as the default; the maximum seems preferable to the minimum. We should add a test for a smaller size instead; since this is auto-generated by Lua, we can rely on a test for one case to provide information on all similar cases.

cwedgwood: PTAL.

PR #279, opened 2024-01-02
Atari 5200 "Super Cart" target

cwedgwood:
Supports:
 * 32 KiB single bank
 * 128 KiB to 512 KiB "Super Cart" bank-switched ROMs


mos-platform/atari5200-super/clang.cfg
```diff
@@ -0,0 +1,8 @@
+#
+# Allow for 96 usable bytes in the zp, using the same layout and
+# offset as Atari 800 for now to make debugging and analysis easier.
+#
+# Strictly speaking as the lowest 24 are used, we have 256-32-24
+# leaving 200 free.
+-mlto-zp=96
```
asiekierka, line 7: Do we want to keep this?

asiekierka:
Other than that, LGTM

mos-platform/atari5200-super/clang.cfg
cwedgwood: I do for now, but not long-term.

I use the "sameness" to the other platform when looking at hexdumps and disassembly, something mostly familiar.

mos-platform/atari5200-super/init-stack.S
mysterymath, line 1: Instead, merge the common-init-stack library and set the symbol `__stack` to `0x3c00` in the linker script. Saves a file.

mos-platform/atari5200-super/link.ld
```diff
@@ -0,0 +1,278 @@
...
+  32 KiB banks, 1 bank would be a normal 32 KiB cartridge, 128 KiB
+  through 512 KiB would be a Super Cart.
+
+  We have a common "fixed" region that is a prefix for all banks.  The
+  last bank has a tail used to initialize the ROM.
+
+  https://github.com/atari800/atari800/blob/ATARI800_5_0_0/DOC/cart.txt
+*/
+
+/* Default cartridge size.  This isn't the largest possible size to
```
mysterymath, line 12: In the absence of other concerns, this should be the largest size, and tests should be provided for any other sizes that might break for independent reasons.

mos-platform/atari5200-super/link.ld
```diff
@@ -0,0 +1,278 @@
...
+
+     Cartridge Metadata: 24 bytes
+
+     20 bytes of title at bfe8 (ATASCII)
+     2 year 19xx at bffc - if bffd is 0xff this means diag cart - TODO
+     init at bffe
+
+     TODO: We might want to juse use things in diag cart mode (some
+     games do this) and ignore the above.  */
+  .vector __tail_lma + (64 - 24) : {
```
mysterymath, line 223: Could we make these configurable symbols with PROVIDEed values? This is used to great effect in the NES setup.

Seems like this would work for everything except the title; if we wanted that to be configurable (which seems desirable), we'd probably need to make it a weak symbol in a retained section. The user could override the symbol and provide their own 20 byte section to replace the one available in atari2500-super's libcrt.

We should probably do the low hanging fruit as part of this PR (any numeric metadata), but NAR on the ASCII; that would require some experimentation.

mos-platform/atari5200-super/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+# We are NOT using atari8-common as it's sufficiently at present,
```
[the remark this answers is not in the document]
cwedgwood: I'll remove it.  The comment was really more something for myself, as to why it doesn't use atari8-common as the parent (the differences right now are too great).

mos-platform/atari5200-super/link.ld
cwedgwood: Will do.

mos-platform/atari5200-super/init-stack.S
cwedgwood: PTAL.

mos-platform/atari5200-super/link.ld
cwedgwood: @mysterymath How allowing a quirky replacement as I have it now?

I agree some mechanism to see the string would be useful, but it also might make sense later on to replace the init code entirely with something we provide, in which case we can do whatever we want in-terms of boot logo, year (at the cost of code-size).

mysterymath (approved):
LGTM

PR #280, opened 2024-01-02
fix nes-nrom "no-compile" tests to fail for the right reasons

cwedgwood:
nrom doesn't have mapper.h, use ines.h


asiekierka:
LGTM

PR #293, opened 2024-01-27
Picocomputer 6502 (RP6502), add new target

rumbledethumps:
Support for: https://picocomputer.github.io/
OK to squash. This is everything not related to the filesystem.


mos-platform/rp6502/rp6502.h
```diff
@@ -0,0 +1,197 @@
...
+/* XREG location helpers */
+
+#define xreg_ria_keyboard(...) xreg(0, 0, 0, __VA_ARGS__)
+#define xreg_ria_mouse(...) xreg(0, 0, 1, __VA_ARGS__)
+#define xreg_vga_canvas(...) xreg(1, 0, 0, __VA_ARGS__)
+#define xreg_vga_mode(...) xreg(1, 0, 1, __VA_ARGS__)
+
+/* XRAM structure helpers */
+
+#define xram0_struct_set(addr, type, member, val)                              \
```
mysterymath, line 78: This macro and the following should really be functions. If it's desired that they always be inlined, then it's sufficient to mark them with `__attribute__((always_inline))`.

mos-platform/rp6502/init-cpu.s
```diff
@@ -0,0 +1,7 @@
+; Initialize the CPU.
```
mysterymath, line 1: I don't see `_start` referenced anywhere in this PR; what's the platform's startup procedure like?

mos-platform/rp6502/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+
+add_platform_library(rp6502-crt0
+  exit.c
+)
+merge_libraries(rp6502-crt0
+  common-crt0
+  common-init-stack
+  common-copy-zp-data
+  common-zero-bss
+  common-exit-loop
```
mysterymath, line 24: This provides an implementation of `_Exit` that loops, but you've already provided one, so this isn't needed. Instead, include `exit-custom` to pull in an `exit()` implementation that calls your `_Exit()`.

mos-platform/rp6502/exit.c
```diff
@@ -0,0 +1,7 @@
+#include "rp6502.h"
+
+void _exit(int status) {
```
mysterymath, line 3: You'll need to sync to get past the breaking commit d2313256667c80986627440bf42d918b5d7f7f68, which removes POSIX `_exit` in favor of the C standard's `_Exit`.

mysterymath (commented):
Thanks for the PR, this is a very clean implementation (and a very clean architecture!).

mos-platform/rp6502/rp6502.h
rumbledethumps: Because `type` is a struct name and `member` is a struct member name, the best I can do is split this into macros *and* inlines. That makes the code harder to follow with no benefit to the compiled output.

mos-platform/rp6502/init-cpu.s
rumbledethumps: There's no startup other than CLD and TXS for the CPU.

mos-platform/rp6502/CMakeLists.txt
rumbledethumps: done.

mos-platform/rp6502/exit.c
rumbledethumps: done.

mos-platform/rp6502/init-cpu.s
rumbledethumps: It just occurred to me to mention that the 6502 reset vector is 0x0200. No code runs before LLVM-MOS code.

mos-platform/rp6502/rp6502.h
mysterymath: Sorry, missed type; my bad.

mos-platform/rp6502/init-cpu.s
mysterymath: In that case, there should be an assert in the linker script that _start is 0x200. The script as is should guarantee this for the most part, but it should fail if violated. (This could happen if a text segment included data with very high alignment requirements.)

rumbledethumps: I added the origin address and start address to the output_format and changed my tooling to use them.

mysterymath (approved):
Alright, LGTM!

rumbledethumps:
Woohoo, thanks. I'll be making a short video to announce this. Stay tuned.

mysterymath:
> Woohoo, thanks. I'll be making a short video to announce this. Stay tuned.

Awesome! I usually try to cut a release every Monday, so there should be a release with these changes soon.

PR #294, opened 2024-01-30
add all remaining rp6502 OS calls

rumbledethumps:
This completes the entire API for the rp6502 target.  It supports time.h, stdio.h, and some of unistd.h. I thought it might be helpful to have while the standard libraries are being designed and built. 

The OS is based on POSIX so assumptions along those lines should be good. More details are here: https://picocomputer.github.io/api.html

I know it's not completely useful in its current state because many prototypes are missing from the standard library (open, read, etc.). It's not clear how to add prototypes properly without breaking other targets.  Users know this is all experimental so breaking changes are no problem.


mysterymath:
> I know it's not completely useful in its current state because many prototypes are missing from the standard library (open, read, etc.). It's not clear how to add prototypes properly without breaking other targets. Users know this is all experimental so breaking changes are no problem.

Could these things go into `rp6502.h` for now?

mos-platform/rp6502/clock.h
```diff
@@ -0,0 +1,16 @@
+#ifndef _CLOCK_H
+#define _CLOCK_H
+
+struct timespec {
```
mysterymath, line 4: This will definitely end up conflicting with `<time.h>`, so if you're up for it, this seems like a good time to establish a minimum `<time.h>` in `common` with just this stuff in it.

mos-platform/rp6502/clock.h
```diff
@@ -0,0 +1,16 @@
...
+  long tv_sec;
```
mysterymath, line 5: This should be `time_t tv_sec`.

Also, I'll probably get flak for this, but 2038 is only 14 years away, and I do expect this project to still be around by then. Given that, and how rarely time manipulation will probably end up being a performance bottleneck, it really seems like `time_t` should be `long long`.

Notably, if you don't feel the same about rp6502, then it would probably be fine to upcast from its internal 32-bit time to a 64-bit `time_t`.

mysterymath (approved):

mos-platform/rp6502/clock.h
rumbledethumps: The RP6502 OS uses a long long internally, but cc65 doesn't support long long. I can make new OS calls that support long long (but not now). There's no hurry, the current ones promote fine, and it can be done in a non-breaking way at any time.

I populated common time.h with a long long time.h and removed clock.h.

I also populated rp6502.h with temporary stdio things. Feel free to delete these as needed.

mysterymath:
Alright, LGTM!

PR #300, opened 2024-02-10
Atari' cartridge target renaming and documentation cleanups

cwedgwood:


mos-platform/CMakeLists.txt
```diff
@@ -49,12 +49,12 @@ add_subdirectory(common)
 add_subdirectory(atari2600-common)
 add_subdirectory(atari2600-4k)
 add_subdirectory(atari2600-3e)
-add_subdirectory(atari5200-super)
+add_subdirectory(atari5200-supercart)
```
mysterymath, line 52: It doesn't seem worthwhile to call out "cart" for targets that have no other way to supply binaries. The benefit that consistency is uniform treatment of indiscernables, but it doesn't seem likely for confusion to occur between targets that support multiple input types and catridge-only machines.

mos-platform/CMakeLists.txt
```diff
@@ -49,12 +49,12 @@ add_subdirectory(common)
...
 add_subdirectory(atari8-common)
 add_subdirectory(atari8-dos)
 add_subdirectory(atari8-stdcart)
```
mysterymath, line 55: Instead, it seems better to break out a term in the hierarchy: `atari8-cart-xegs`. This provide better parity with `atari8-dos`. It doesn't seem necessary to have the target dash hierarchy match the actual PARENT relation, so long as there isn't actually a `atari8-cart` target.

mos-platform/CMakeLists.txt
mysterymath: Discussed on Discord; this was fine. The canonical name is "Super Cart cartridge", so dropping the improper part of the name does yield "supercart".

README.md
```diff
@@ -9,10 +9,18 @@ The LLVM-MOS compiler toolchain and platform libraries.
 - Atari 2600
   - 4K cartridge
   - [TigerVision 3E (2-KiB fixed, 2-KiB banked ROM or RAM)](https://www.taswegian.com/WoodgrainWizard/tiki-index.php?page=3E)
+- [Atari 5200](https://en.wikipedia.org/wiki/Atari_5200) Super Cart from [32 KiB](https://github.com/atari800/atari800/blob/master/DOC/cart.txt#L105)
```
mysterymath, line 12: These should be GitHub permalinks; the line ordering may yet change. (The last update to `cart.txt` was 9 months ago.)

cwedgwood: Oh, they were supposed to be against a tagged release.  Does that suffice?  People don't usually change tags.

mysterymath: Yeah, that should be fine.

mysterymath (approved):
LGTM!

PR #303, opened 2024-02-11
Add `fmin`/`fmax` functions to `math.h`

### LINKED ISSUE #299: Implement libm min and max functions
In RUST cast from float to int (with `as` keyword) is saturating operation: https://doc.rust-lang.org/reference/expressions/operator-expr.html#numeric-cast

and for now it generates following error:

```
LLVM ERROR: unable to legalize instruction: %22:_(s32) = G_INTRINSIC intrinsic(@llvm.fptosi.sat), %8:_(s32) (in function: main)
```

mlund:
This adds min and max functions for `float` and `double`.
Closes #299.


mos-platform/common/c/math.cc
```diff
@@ -0,0 +1,22 @@
+#include <math.h>
+
+template <typename T> static inline bool _is_nan(const T x) { return x != x; }
```
mysterymath, line 3: May as well actually implement the macro in `<math.h>`:

`#define isnan(arg) __builtin_isfpclass(arg, __FPCLASS_SNAN | __FPCLASS_QNAN)`

mos-platform/common/c/math.cc
```diff
@@ -0,0 +1,22 @@
...
+
+template <typename T> static inline T _fmax(const T x, const T y) {
+  if (_is_nan(x)) return y;
```
mysterymath, line 6: This isn't needed; if `x` is nan, then `x > y` is false, so the ternary returns `y`. Replace this line with a comment to this effect.

mos-platform/common/c/math.cc
```diff
@@ -0,0 +1,22 @@
...
+  if (_is_nan(y)) return x;
+  return x > y ? x : y;
+}
+
+template <typename T> static inline T _fmin(const T x, const T y) {
+  if (_is_nan(x)) return y;
```
mysterymath, line 12: Same here.

mos-platform/common/c/math.cc
mlund: Nice catch - fixed.

mos-platform/common/c/math.cc
mlund: fixed

mos-platform/common/c/math.cc
mlund: Or use the existing `__builtin_isnan`?

mlund: Also, I now notice that `__builtin_fmin` and `__builtin_fmax` along with several other math functions are already defined and seem to be correct. Perhaps better to use these? Are these covered by existing tests in the separate test suite?

mysterymath: Yeah, `__builtin_isnan` lowers to the same thing and is clearer.

`__builtin_fmin` doesn't help us; it lowers to a compiler intrinsic that most naturally lowers to a call to `fmin`. By comparison, `__builtin_isnan` lowers to a call to `__unordsf2` in `compiler_rt`, plus a bit of additional logic.

mysterymath (approved):
LGTM!

PR #306, opened 2024-02-17
Make __heap_start relocatable rather than absolute.

davidgiven:
In the GNU linker language, `ALIGN(alignment)` actually produces an absolute value, while `ALIGN(address, alignment)` produces a relocatable value. See https://sourceware.org/binutils/docs/ld/Builtin-Functions.html#index-ALIGN_0028align_0029 which states that the first is equivalent to `ALIGN(ABSOLUTE(address), alignment)`. Wheee!

This causes `__heap_start` on CP/M-65 to be an absolute value rather than a relocatable one, _even though it's defined in a relocatable segment_, which meant the heap address wasn't getting relocated to the right place, causing big chunks of the program to be overwritten --- with different results on each target platform because they each loaded binaries in different places...

### DIFF
--- a/mos-platform/common/ldscripts/noinit-sections.ld
+++ b/mos-platform/common/ldscripts/noinit-sections.ld
@@ -1,3 +1,3 @@
 *(.noinit .noinit.* NULL INIT ZPSAVE)
 /* This helps ensure all chunk sizes have the low bit free for use as a flag. */
-__heap_start = ALIGN(2);
+__heap_start = ALIGN(., 2);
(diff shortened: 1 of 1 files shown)


mos-platform/common/ldscripts/noinit-sections.ld
```diff
@@ -1,3 +1,3 @@
 *(.noinit .noinit.* NULL INIT ZPSAVE)
 /* This helps ensure all chunk sizes have the low bit free for use as a flag. */
-__heap_start = ALIGN(2);
+__heap_start = ALIGN(., 2);
```
mysterymath, line 3: Great find; definitely add a comment here though that CP/M-65 depends on this, and why. Otherwise, anyone seeing this would have to find this PR via git blame.

mysterymath (approved):

mos-platform/common/ldscripts/noinit-sections.ld
davidgiven: Done. This is, AFAICT, the only place in the common code where ALIGN is used like this. I don't know if any of the other platforms have relocatable binaries (has someone approached you about O65 binaries?) but I believe this is the only instance affecting me.

PR #312, opened 2024-02-18
[Breaking] Rename atari8-cart' targets to be simple triples

cwedgwood:
For some reason the upstream changed how it treats triples so that anything after the triple is now consider a version.  Rather than buck this, go with it for now.

Fixes: #311


cwedgwood:
@mysterymath RFC/thoughts, nothing pressing right now.

mysterymath:
There's nothing blessed about these triples, it just happens to be what the frontend currently accepts. If it doesn't universally accept the names we give it, then we need to either understand the intent behind its existing organization scheme, and conform our names to that, or we need to broaden what the frontend is willing to accept. This change does neither, so I'm closing it. We'll need to resolve this before the next release, but there's no need to be hasty.

PR #316, opened 2024-02-24
Add fixed point math library

jroweboy:
Add a general purpose fixed point math library. Leaving it as a draft PR until I get the time to add the test cases to llvm-test-suite and fix any issues we find there. (I wasn't able to build locally so I've been developing in godbolt in the meantime)

Also I think there's some ambiguous resolve errors for comparing fixed point numbers right now that I need to fix before merging too. I would expect that `number = 9.4_8_8; number > 7.7_8_8` should work, but it can't decide which comparison operator to use and errors out currently.

I am just pushing this here so I don't lose track of it, but if you want to do an early review to help out with code clarity/style/etc feel free to do so!


mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
+
+
+#ifndef _FIXED_POINT_H
+#define _FIXED_POINT_H
+
+// Copyright 2024 LLVM-MOS Project
```
mysterymath, line 6: Nit; the license header should be at the very start of the file; it applies to everything, including the include guard ;)

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+/// With our 8.8 example, the Fractional portion can represent 256 values total,
+/// meaning the value in the fractional part represents a number divided by 256.
+/// As a practical example, a 16 bit number `0x5060` , and when divided by `0x100`
+/// yields `0x50` with a remainder of `0x60`. This class makes that divider a
+/// part of the type of the class, allowing seamless math operations as if its a
+/// normal integer, but with the correct shifting when combining with other values.
+template<uint8_t ISize, uint8_t FSize, bool Signed = true> class Value;
+
+template<uint8_t ISize, uint8_t FSize, bool Signed>
+class Value {
```
mysterymath, line 45: `fixedpoint::Value` is a bit of a mouthful, and `using namespace fixpoint;` would leave just `Value`, which is to ambiguous to use.

Instead, I'd recommend placing the class and the typedefs in the global namespace, and using namespace `fixedpoint_literals` for the literals. `using namespace fixedpoint_literals;` isn't too much typing, since it's once per file, and it would allow referring to arbitrary types as `FixedPoint<12,34> x;`.

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+  template <typename R>
+  static constexpr R BitMask(unsigned int const onecount) {
+    return static_cast<R>(-(onecount != 0))
+      & (static_cast<R>(-1) >> ((sizeof(R) * CHAR_BIT) - onecount));
+  }
+
+  // Size of the underlying storage for the different views of the data.
+  // The total combined storage will be used for representing the actual value.
+  using IntType = std::conditional_t<Signed, 
+                                     signed _BitInt(((ISize + 7) / 8) * 8),
```
mysterymath, line 58: Extract the widths out into named constexpr constants; the overall logic here gets hard to read out of all the arithmetic.

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+/// yields `0x50` with a remainder of `0x60`. This class makes that divider a
+/// part of the type of the class, allowing seamless math operations as if its a
+/// normal integer, but with the correct shifting when combining with other values.
+template<uint8_t ISize, uint8_t FSize, bool Signed = true> class Value;
+
+template<uint8_t ISize, uint8_t FSize, bool Signed>
+class Value {
+protected:
+
+  // Used to calculate the bitmasks for both the integer part and fractional part
```
mysterymath, line 48: Can you provide more detail about the semantics of onecount and the return value?

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+  template <typename R>
+  static constexpr R BitMask(unsigned int const onecount) {
+    return static_cast<R>(-(onecount != 0))
+      & (static_cast<R>(-1) >> ((sizeof(R) * CHAR_BIT) - onecount));
+  }
+
+  // Size of the underlying storage for the different views of the data.
+  // The total combined storage will be used for representing the actual value.
+  using IntType = std::conditional_t<Signed,
```
mysterymath, line 57: Since they're used in public methods, they should be public, since it's otherwise impossible to name them without `auto`.

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+/// normal integer, but with the correct shifting when combining with other values.
+template<uint8_t ISize, uint8_t FSize, bool Signed = true> class Value;
+
+template<uint8_t ISize, uint8_t FSize, bool Signed>
+class Value {
+protected:
+
+  // Used to calculate the bitmasks for both the integer part and fractional part
+  template <typename R>
+  static constexpr R BitMask(unsigned int const onecount) {
```
mysterymath, line 50: nit: `bitMask`

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+/// meaning the value in the fractional part represents a number divided by 256.
+/// As a practical example, a 16 bit number `0x5060` , and when divided by `0x100`
+/// yields `0x50` with a remainder of `0x60`. This class makes that divider a
+/// part of the type of the class, allowing seamless math operations as if its a
+/// normal integer, but with the correct shifting when combining with other values.
+template<uint8_t ISize, uint8_t FSize, bool Signed = true> class Value;
+
+template<uint8_t ISize, uint8_t FSize, bool Signed>
+class Value {
+protected:
```
mysterymath, line 46: Protected interfaces form another public API surface; it's just between a class and its descendants, rather than a class and it's users. Accordingly, making every field protected is morally equivalent to making every field public; encapsulation would tend to suggest that everything be private until the purpose of a protected API reveals itself.

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+      & (static_cast<R>(-1) >> ((sizeof(R) * CHAR_BIT) - onecount));
+  }
+
+  // Size of the underlying storage for the different views of the data.
+  // The total combined storage will be used for representing the actual value.
+  using IntType = std::conditional_t<Signed, 
+                                     signed _BitInt(((ISize + 7) / 8) * 8),
+                                     unsigned _BitInt(((ISize + 7) / 8) * 8)>;
+  using FracType = std::conditional_t<Signed,
+                                      signed _BitInt(((FSize + 7) / 8) * 8),
```
mysterymath, line 61: Alternatively to the above, consider extracting out the _BitInt into a typedef and using std::make_unsigned on it (assuming that works? unsure.)

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+                                      unsigned _BitInt(((FSize + 7) / 8) * 8)>;
+  using StorageType = std::conditional_t<Signed,
+                                         signed _BitInt((((ISize + FSize) + 7) / 8) * 8),
+                                         unsigned _BitInt((((ISize + FSize) + 7) / 8) * 8)>;
+  StorageType val;
+  static constexpr unsigned long INT_MASK = BitMask<unsigned long>(ISize) << FSize;
```
mysterymath, line 67: `unsigned long` doesn't seem long enough to hold these in the general case. We should either use some kind of variable, responsive type or place limits on the legal values of the template parameters.

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+  static constexpr unsigned long FRAC_MASK = BitMask<unsigned long>(FSize);
+
+public:
+
+  // Constructors
+
+  /// Common constructor for creating a fixed point value from an integer value
+  [[clang::always_inline]] constexpr explicit Value(IntType i) {
```
mysterymath, line 75: The general zeitgeist in C++ is to have lossless, cheap conversions be implicit. This appears to be the case for `IntType` -> `FixedPoint`. Having this implicit would allow you to e.g. pass an integer to a function that expects a fixed point; compare passing an integer to a function that expects a float.

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+    SetI(i); 
+  }
+  
+  /// Copy constructor for a fixed point value with different sizes of Int/Frac
+  template<uint8_t OI, uint8_t OF, bool S>
+  [[clang::always_inline]] constexpr Value(const Value<OI, OF, S>& o) {
```
mysterymath, line 81: Similar to the above, this should be implicit if `OI >= ISize && OF >= FSize` (i.e. lossless), and explicit otherwise (lossy). I think some `std::enable_if`-fu can express this; may take some experimentation.

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+namespace literals {
+
+using FP_S8_8 = Value<8, 8, true>;
+using FP_S12_4 = Value<12, 4, true>;
+using FP_U8_8 = Value<8, 8, false>;
+using FP_U12_4 = Value<12, 4, false>;
+using FP_8_8 = FP_S8_8;
+using FP_12_4 = FP_S12_4;
+
+[[clang::always_inline]] constexpr FP_S8_8 operator ""_s8_8(long double fixed) {
```
mysterymath, line 278: Add a preprocessor check for C++20.

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+  /// Copy constructor for fixed point values with the same size
+  [[clang::always_inline]] constexpr Value(const Value& o) : val(o.val) {}
+  
+  /// Constructor for setting both the integral and fractional part
+  [[clang::always_inline]] constexpr explicit Value(IntType i, FracType f) {
+    SetI(i);
+    SetF(f);
+  }
+  /// Constructor to convert floating point values into fixed point
+  [[clang::always_inline]] constexpr Value(long double f) {
```
mysterymath, line 93: This case is trickier; converting a long double to a fixed point number may or may not be lossy. It'll almost always be lossy though. And at runtime, it will be quite slow.

If we have C++20 support (preprocessor check), it seems like these constructors should be `consteval` to ensure that they can only be evaluated at compile time. (I *think* `consteval` constructors work this way; if it doesn't , NAR.) In either case, we'd probably want non-constexpr versions of this for the three float types.

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+    Set(f * (1 << FSize));
```
mysterymath, line 94: `// Multiplying by a power of two only increases the exponent of the floating point number, which is exact so long as overflow does not occur. Rounding a floating point number to whole is always exact.`

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+public:
+
+  // Constructors
+
+  /// Common constructor for creating a fixed point value from an integer value
+  [[clang::always_inline]] constexpr explicit Value(IntType i) {
+    SetI(i); 
+  }
+  
+  /// Copy constructor for a fixed point value with different sizes of Int/Frac
```
mysterymath, line 79: The various lossy conversion functions below need descriptions of what inputs are exactly representable, and if not exactly representable, either a description of the behavior, or a description of which input and output combinations produce undefined behavior.

There's a choice point here; truncation of fractional bits should definitely be legal, but it could either round to even like floating point or truncate like integers. Given that this is intended to be an integer-based approximation to floating point for highest speed, I'd assume the latter, but fixed point libraries I've seen in the wild do the former. In either case, that behavior should be documented.

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+  }
+  /// Constructor to convert floating point values into fixed point
+  [[clang::always_inline]] constexpr Value(long double f) {
+    Set(f * (1 << FSize));
+  }
+
+  // Direct value accessor and setter methods
+
+  /// Returns just the integral portion
+  [[clang::always_inline]] constexpr IntType AsI() const { return (val & INT_MASK) >> FSize; }
```
mysterymath, line 100: We've been using snake_case for variables and functions; so `as_i()` etc.

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+  /// Returns just the fractional portion
+  [[clang::always_inline]] constexpr FracType AsF() const { return (val & FRAC_MASK); }
+  /// Returns the entire value
+  [[clang::always_inline]] constexpr StorageType Get() const { return val; }
+  /// Update just the integral portion
+  [[clang::always_inline]] constexpr void SetI(IntType v) { 
+    val = (val & ~INT_MASK) | static_cast<StorageType>((v) & (INT_MASK >> FSize)) << FSize;
```
mysterymath, line 107: This kind of logic seems a lot easier to express using bit fields in a struct; does that also produce sufficient codegen quality?

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+  /// Update the entire value
+  [[clang::always_inline]] constexpr void Set(StorageType v) { val = v; }
+
+  // Operator overloads
+  [[clang::always_inline]] constexpr Value& operator=(Value o) {
+    val = o.val;
+    return *this;
+  }
+
+  [[clang::always_inline]] constexpr Value operator +(const Value& o) const {
```
mysterymath, line 122: Similarly to the above, seems most sensible to take `o` by value, not by constant reference. Here and throughout.

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+    Value<ISize, FSize> n = val;
```
mysterymath, line 123: This doesn't seem right; there isn't a constructor that takes in storage type.
Maybe `Value n = *this;` instead? Here and throughout.

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+    n -= o;
+    return n;
+  }
+  [[clang::always_inline]] constexpr Value& operator -=(const Value& o) {
+    val -= o.val;
+    return *this;
+  }
+
+  template<typename T, typename = typename std::enable_if<std::is_arithmetic<T>::value, T>::type>
```
mysterymath, line 142: The `enable_if` is superfluous here and below; the template will already only expand for types that we can divide the value by. Those are exactly the types that it's meaningful to implement division for.

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+    n <<= v;
+    return n;
+  }
+  template<typename T, typename = typename std::enable_if<std::is_arithmetic<T>::value, T>::type>
+  [[clang::always_inline]] constexpr Value& operator <<=(T v) {
+    val <<= v;
+    return *this;
+  }
+
+  [[clang::always_inline]] Value &operator++() {
```
mysterymath, line 190: Hmm, I'd think that this operation would increment the integer part, not the base representation. It should be logically equivalent to `+= 1`, and right now it isn't.

It might be okay to have a break here if there's a compelling reason, but that reason should definitely be written down somewhere at the very least.

mos-platform/common/include/fixed_point.h
```diff
@@ -0,0 +1,301 @@
...
+  }
+  [[clang::always_inline]] constexpr friend bool operator>=(const Value& a, const Value& b) noexcept {
+    return !(a < b);
+  }
+  [[clang::always_inline]] constexpr friend bool operator<=(const Value& a, const Value& b) noexcept {
+    return !(b < a);
+  }
+
+  [[clang::always_inline]] constexpr friend bool operator==(const Value& a, IntType b) noexcept {
+    return a.val == b;
```
mysterymath, line 222: These aren't correct; they don't mask away the fractional bits or shift the integer to put it in line with the integer bits.

johnwbyrd:
There's a great deal of relevant prior art here.  Examples:

https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0037r7.html
https://www.open-std.org/JTC1/SC22/WG14/www/docs/n1169.pdf
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0106r0.html
https://github.com/mizvekov/fp

- It looks to me that your code rounds the int and frac components to byte boundaries.  What about the case of odd fractional formats, such as 10.8 or 5.3?  I don't see your code exercising those cases.

- I don't see code for converting between various frac formats, such as 8.8 to 12.4 and the like.  Perhaps because of that...

- I don't see support for multiplication or division of your novel type.  Handling the various cases in multiplication and division (choosing a destination precision, respecting or disrespecting overflows, respecting the sign bit if any, normalizing to the destination type) is one of the most inconvenient problems in computing with frac's, and I would hope that we could help out the user in this regard.

Have courage...  I don't believe that you are solving an easy problem.

mysterymath:
> * I don't see support for multiplication or division of your novel type.  Handling the various cases in multiplication and division (choosing a destination precision, respecting or disrespecting overflows, respecting the sign bit if any, normalizing to the destination type) is one of the most inconvenient problems in computing with frac's, and I would hope that we could help out the user in this regard.

I'd like to push back on this, at least for multiplication, since implementing it well is quite nontrivial. I don't think an initial version of this needs to support it, since it's also not very common in the NES scene that is requesting the feature to multiply fixed points together (correct me if I'm wrong on this one).

johnwbyrd:
> Handling the various cases in multiplication and  implementing it well is quite nontrivial. I don't think an initial version of this needs to support it, since it's also not very common in the NES scene that is requesting the feature

I think it's worth considering now, exactly because of its nontriviality.  It is exactly the sort of thing that's hard to add later, once you bet the farm on a specific architecture.  At the very least, I think it's not unreasonable to ask @jroweboy to review the references that I provided, and report back and see if he's had any novel enlightenment.  Even if we don't implement it now, I'd like to be assured that we *can* implement it later.

mysterymath:
> > Handling the various cases in multiplication and  implementing it well is quite nontrivial. I don't think an initial version of this needs to support it, since it's also not very common in the NES scene that is requesting the feature
> 
> I think it's worth considering now, exactly because of its nontriviality. It is exactly the sort of thing that's hard to add later, once you bet the farm on a specific architecture. At the very least, I think it's not unreasonable to ask @jroweboy to review the references that I provided, and report back and see if he's had any novel enlightenment. Even if we don't implement it now, I'd like to be assured that we _can_ implement it later.

It's not hard to add later; we still have very loose backwards compatibility guarantees, and the newer the feature, the looser they are. Better is the enemy of good.

johnwbyrd:
The NES community has waited approximately 39 years for this feature; I don't necessarily think a day or two of delay will break anyone's heart.

I'm not saying don't do it; I'm just saying that now is the best time for due diligence and stealing good ideas.

mysterymath:
> The NES community has waited approximately 39 years for this feature; I don't necessarily think a day or two of delay will break anyone's heart.
> 
> I'm not saying don't do it; I'm just saying that now is the best time for due diligence and stealing good ideas.

I don't think that literature reviews should be a prerequisite for contributing to a hobby compiler that we work on for fun. llvm-mos should welcome authors with all degrees of seriousness, and the purpose of asking more from someone in a code review should only to be ensure that the fun the author is having is balanced well with the fun the users of what they write will have.

In any case, I don't think this PR is the correct venue for this discussion; I'd suggest moving this to Discord. In the interim, I'll use my sway as current maintainer to say that this PR should proceed as-is if and while we have a broader discussion about what we want our bar to be.

jroweboy:
> It looks to me that your code rounds the int and frac components to byte boundaries. What about the case of odd fractional formats, such as 10.8 or 5.3? I don't see your code exercising those cases.

Thanks to feedback here, I'm testing out using the bitfield support built-in to the compiler to see if the code gen is still just as good. Note that even in this original code it still supports 5.3 as the `StorageSize` is just the sum of the int/frac types, and masking is used to break it down into the correct int/frac sizes. The `IntSize` being rounded up was just to get better codegen since it didn't like `as_i` and other methods to take a non-rounded bitint number.

> There's a great deal of relevant prior art here

Yes! And there's also prior art for high level languages with the 6502 as well. I spoke with pubby (maker of nesfab) which is a compiler with builtin fixed point support to get some ideas of what is expected from this feature. NESFab only support byte sized fractional values, but still has compile time float -> fixed support like I want to offer here. Multiplication extends the values into the combined size of the output type (so if you have an 8.8 * 12.4 it becomes 20.12) and division is unimplemented as its rather troublesome for something people don't use in practice. 

> The test suite (if that was its intention) seems a little thin.

I mentioned this in the original post, but the test suite will be added to llvm-mos-lit test cases once I've established the implementation a little better. 


I'm down to work towards a multiplication solution that addresses all of your concerns, especially since there's existing art for this on 6502 systems (well at least the one i'm familiar with), but I'm not so keen on division as it quickly becomes problematic for very little benefit IMO.

mysterymath:
> I'm down to work towards a multiplication solution that addresses all of your concerns, especially since there's existing art for this on 6502 systems (well at least the one i'm familiar with), but I'm not so keen on division as it quickly becomes problematic for very little benefit IMO.

So long as the fixed point type sizes are identical, `(a * 2^-f) / (b * 2^-f) = a / b = (a * 2^f / b) * 2^-f. Accordingly, it's only slightly more difficult than division by a non-fixed-point integer: `a * 2^-f / b = (a / b) * 2^-f`. It gets complex if the fixed point fractions are different, but that can be also be called out as a compile time error. The difficulty in this case, as with multiplication, is that wider arithimatic must be done, and it can be difficult to do this efficiently.

PR #323, opened 2024-03-10
[neo6502] New port.

asiekierka:
Adds support for OLIMEX's [Neo6502 board](https://github.com/OLIMEX/Neo6502).

The original `neo6502.inc` is adapted and included; as well as a full C adaptation in `neo6502.h` and `kernel.h`. In addition, `clock`/`getchar`/`putchar` is implemented.


mos-platform/neo6502/api/console.c
```diff
@@ -0,0 +1,109 @@
...
+// Licensed under the Apache License, Version 2.0 with LLVM Exceptions.
+// See https://github.com/llvm-mos/llvm-mos-sdk/blob/main/LICENSE for license
+// information.
+
+#include <string.h>
+#include "../neo6502.h"
+#include "../kernel.h"
+#include "api-internal.h"
+
+__attribute__((leaf))
```
mysterymath, line 11: Leaf is useless on definitions.

mos-platform/neo6502/api/api-internal.h
```diff
@@ -0,0 +1,26 @@
...
+#include <stdint.h>
+#include <string.h>
+
+#ifndef _NEO_API_INTERNAL_H
+#define _NEO_API_INTERNAL_H
+
+/**
+ * @brief Given an input C-format string FROM, provide a Pascal-format string FROM_p using VLAs.
+ */
+#define PASCALIZE_INPUT(FROM) \
```
mysterymath, line 15: It's considered good macro practice (coming from the LISP world) to have macros that bind new public names do so by taking the name as an argument. Even though it can always be advertised as a formula derived from the given name, it requires less knowledge of the macro to read `PASCALIZE_INPUT(str_p, str); oh that must have defined this reference to str_p;` than `PASCALIZE_INPUT(str); str_p appears outta nowhere`.

mos-platform/neo6502/api/console.c
```diff
@@ -0,0 +1,109 @@
...
+void neo_console_read_line(char *line) {
+    line[0] = 0;
+    neo_console_read_line_p(line);
+    __neo_depascalize_output(line);
+}
+
+__attribute__((leaf))
+void neo_console_define_hotkey_p(uint8_t hotkey, const void *str) {
```
mysterymath, line 37: Was there a reason that the strings in the neo6502 APIs were declared `void*` instead of `char*`?

mos-platform/neo6502/api/neo/console.h
```diff
@@ -0,0 +1,151 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief Read and remove a key press.
+ *
+ * @return ASCII key value; 0 if no key presses queued.
+ */
+__attribute__((leaf))
```
mysterymath, line 22: This attribute doesn't do anything on functions where definitions are provided in the LTO unilt.

mysterymath (approved):

mos-platform/neo6502/api/console.c
asiekierka: Sorry, I kind of grew the habit of putting the attribute in too many places as opposed to too few. Thanks!

mos-platform/neo6502/api/console.c
asiekierka: I wanted to quickly distinguish the ones which take Pascal string types from the ones which take C string types. Not sure if I should just have them as `const char *` or if I should define a `struct neo_pstring` or...

asiekierka:
Should be fixed, now. Thank you!

PR #324, opened 2024-03-10
[supervision] New port.

asiekierka:
Port for the Watara Supervision.

I have chosen not to copy the cc65 header, as further discoveries to the hardware had been made since it was written; where applicable, I have followed some of its conventions, however.

Some basic utility functions have also been implemented.


mysterymath (approved):
LGTM!

PR #329, opened 2024-03-27
[lynx] Initial target support

asiekierka:
This implements support for outputting Atari Lynx BLL (`.bll.o`) executables, as well as a "common" target providing cc65's hardware definition headers.

This is a "bare minimum" PR, but can be expanded upon later by future PRs.


mysterymath (approved):
LGTM

PR #332, opened 2024-05-28
[a8 dos] Exit via DOSVEC

cwedgwood:


mysterymath (commented):
There's nothing unusual about `JMP ($a)` as an implementation of `_Exit()` (the C standard function that performs an immediate exit with no cleanup). Accordingly, we should merge the library `exit-custom` from `common` to provide the full `exit()` and after-main functionality in terms of `_Exit()` in `atari-dos`'s `crt0`, and then the `atari-dos` target only needs to implement `_Exit` in its own `_Exit.s`

mysterymath:
Went ahead and fixed this; landing.

PR #337, opened 2024-06-14
[atari] Optimize findfreeiocb

pfusik:


mos-platform/atari8-common/findfreeiocb.s
```diff
@@ -16,16 +16,11 @@
...
         beq     found
-        txa
-        clc
-        adc     #$10
-        tax
-        cmp     #$80
-        bcc     loop
-        inx                     ; return ZF cleared
+        adc     #$10            ; C = 0
+        bpl     loop
```
mysterymath, line 25: This should have a comment explaining why `bpl` gives the loop termination condition, now that it's no longer an explicit compare against `#$80`.

pfusik: What's your point? Checking the N flag is the 6502's idiomatic way to determine $00-$7F vs $80-$FF. It's the `cmp #$80` that looked strange.
Also, all Atari programmers should be aware that there are 8 IOCBs.
I'm okay with adding a comment, but I don't know what should it explain?

mysterymath: It's not actually checking whether the value is negative, it's literally checking whether it is >= 0x80. It's looping through the IOCBs, each of which are 0x10 bytes long, and there are 8 of them.

mysterymath (approved):
LGTM!

PR #340, opened 2024-06-17
[atari] Optimize fd routines

pfusik:


mos-platform/atari8-common/getfd.s
```diff
@@ -48,15 +47,14 @@ found:  tya
...
+; registers destroyed
 getfd:
 
-        sta     __rc10          ; save fd_table entry
         jsr     fdt_to_fdi
         bcs     error
 
         pha
-        lda     __rc10
+        tya
```
mysterymath, line 57: The value in Y isn't part of the listed contract for `fdt_to_fdi`, so the invariant needs to be added to the header comment to prevent it from being accidentally broken by changes to that routine.

mysterymath (approved):
LGTM!

Issue #341, opened 2024-06-25
`commodore` family targets shouldn't both clobber BASIC ZP and return to BASIC


mysterymath:
This came up in https://github.com/llvm-mos/llvm-mos-sdk/pull/315, and it's discussed more thoroughly in the tail end of that discussion.

Summary:
I think there's three aspects of the commodore family targets that come into conflict:
- That the only real execution environment that loads programs is BASIC
- That they exit by returning (from the above, always to BASIC)
- That they clobber BASIC ZP.
If any of the above weren't true, there'd be a way to use the resulting PRG files such that they cleanly returned, but as is it feels like one of the second two points should change. We may want to either split the commodore targets into BASIC and no-BASIC variants, or if it's possible, provide a link-time configuration symbol for this. Whatever the default ends up being, it should infinite loop if it isn't safe to return to BASIC; that at least may let one inspect the output.

The easiest change we could make to make this consistent would be to change the commodore targets to exit-loop instead of exit-return; that's one line of CMake. It looks like cc65 doesn't use the BASIC ZP at all and returns. That's one possible configuration. Another is to save and restore BASIC ZP on entry/exit and use it, and another is to clobber it and never return. Ideally there would be a simple way to spell each of these for each commodore target in the SDK, and hopefully, without too much monkeying around in linker scripts. That's the part that needs designing; we've built out a huge degree of this kind of configuration in the NES targets, so it may be possible, but it can be tricky.

mysterymath:
I think saving and restoring BASIC ZP makes sense as a library actually. We already have symbolic names for the BASIC ZP area, so we could register actions in early init/late fini to save and restore to a buffer. The buffer would need to be defined in a `.s` file per target (different sizes), but that's no major issue either.

mysterymath:
Closing this out with #342 and #343.

PR #342, opened 2024-06-25
[Breaking] Make commodore family targets infinitely loop on exit

mysterymath:
We currently take advantage of the BASIC ZP area, but commodore family
targets can only load programs from a BASIC environment. This makes it
intrinsically unsafe to exit by return, since it always returns to a
possibly-clobbered environment.

Instead, exiting by looping still allows a screen's worth of output to
be viewed, without possibly returning to a broken BASIC interpreter.
This is arguably a bugfix, but it's also a very significant change in
behavior, so I'm considering it a breaking change.

We should also have an optional library that saves and restores BASIC ZP
on entry-exit and exits via returning, but that shouldn't block this
change.


mysterymath:
See #341

PR #347, opened 2024-07-14
[mega65] Add DMA hardware registers and DMA examples

mlund:
The following adds support for the DMA controller on the MEGA65.
- [x] Add `_dmagic.h` C header
- [x] Add volatile MMIO access to F018 "DMAgic" controller
- [x] Add C++ support library for constructing arbitrary DMA jobs as well as fill and copy functions.
- [x] Add minimal C++ example with DMA fill and copy on screen. ~~There is still an issue with DMA fill/copy routines likely due to code being optimized away. Will try to solve this on Discord.~~
- [x] Add C++ example of vertical raster bars using the DMA hold feature as described in the MEGA65 Book. Below is a screenshot of the effect.
vertical_raster
- [x] Add DMA audio example for playing digitized sound samples. Preview:
https://github.com/user-attachments/assets/8b6207fb-d81e-4daa-86e5-a050f1df026e


mos-platform/mega65/dma.hpp
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Perform enhanced DMA action defined in DMAJob structure.
+ */
+template <size_t N, typename T>
+inline void trigger_dma(const DMAJob<N, T> &dma_job) {
+  DMA.enable_f018b = std::is_same<T, DMAList_F018B>::value;
+  DMA.addr_bank = 0;
+  DMA.addr_msb = ((uint16_t)&dma_job) >> 8;
+  DMA.trigger_enhanced = ((uint16_t)&dma_job) & 0xff;
+  // The following is to prevent `dma_job` from being optimized out
```
mlund, line 97: @mysterymath So far this is the only way I managed to prevent this function from being optimized out. Combination of volatile and pointers unfortunately didn't seem to work. This solution seems like quite the hack and any input is most welcome.

mlund: Following our discussion on github, I conclude that `asm volatile` is an appropriate solution and will leave as is.

examples/mega65/vertical_raster.cc
```diff
@@ -0,0 +1,88 @@
...
+  // Set custom color palette
+  for (size_t i = 1; i < 16; i++) {
+    PALETTE.red[i] = 15 - i;
+    PALETTE.green[i] = 0;
+    PALETTE.blue[i] = i;
+  }
+
+  // Configure DMA job; see The MEGA65 Book, Appendix O.
+  dma::DMAJob<2, DMAList_F018A> dma;
+  dma.options[0] = DST_ADDR_BITS_OPT;
```
mlund, line 66: Here an initializer list would have been nice, but it doesn't seem to be supported by the compiler. That is `dma.options = {DST_ADDR_BITS_OPT, 0xff};`

examples/mega65/CMakeLists.txt
```diff
@@ -4,3 +4,13 @@ install_example(viciv_test.prg)
...
 
+add_executable(vertical_raster.prg vertical_raster.cc)
+install_example(vertical_raster.prg)
+
+add_executable(simple_dma.prg simple_dma.cc)
+install_example(simple_dma.prg)
+
+add_executable(dma_audio.prg dma_audio.c)
+install_example(dma_audio.prg)
+install(FILES drums.s8 TYPE BIN)
```
mlund, line 15: Should `CMAKE_CURRENT_SOURCE_DIR` be added or is that implied? Seems to work correctly as is, though.

mysterymath: This should be fine.

examples/mega65/dma_audio.c
```diff
@@ -0,0 +1,34 @@
...
+#include <mega65.h>
+
+typedef unsigned _BitInt(24) uint24_t;
+
+// Sample data prepared with:
+// ~~~
+// sox -v 2.0 -V -D drums.wav drums.s8 remix - lowpass 7000 rate -v -s -I 11822 dither -S
+// ~~~
+static const uint8_t sample[] = {
+#embed "drums.s8"
```
mysterymath, line 17: NOICE!

examples/mega65/drums.license
```diff
@@ -0,0 +1,4 @@
+Source:      https://commons.wikimedia.org/wiki/File:Patró_de_bateria.wav
+Attribution: Escola Superior de Música de Catalunya (ESMUC)
+License:     CC BY-SA 3.0 <https://creativecommons.org/licenses/by-sa/3.0>, via Wikimedia Commons
```
mysterymath, line 3: This will cause problems. Share-Alike's terms require Adaptations to be licensed under the same license, and the example above is closer to an adaption or derivative work of the drum sample than it is a simple "collection". We'd need to find something else to use either with a compatible license to Apache 2 or in the public domain.

mos-platform/mega65/dma.hpp
```diff
@@ -0,0 +1,100 @@
+// Copyright 2023 LLVM-MOS Project
+// Licensed under the Apache License, Version 2.0 with LLVM Exceptions.
+// See https://github.com/llvm-mos/llvm-mos-sdk/blob/main/LICENSE for license
+// information.
+//
+// C++ support for MEGA65 DMA operations.
+//
+
+#pragma once
```
mysterymath, line 9: Prefer include guards for nit-picky compiler-person reasons.

mos-platform/mega65/dma.hpp
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Perform enhanced DMA action defined in DMAJob structure.
+ */
+template <size_t N, typename T>
+inline void trigger_dma(const DMAJob<N, T> &dma_job) {
+  DMA.enable_f018b = std::is_same<T, DMAList_F018B>::value;
+  DMA.addr_bank = 0;
+  DMA.addr_msb = ((uint16_t)&dma_job) >> 8;
+  DMA.trigger_enhanced = ((uint16_t)&dma_job) & 0xff;
+  asm volatile("");
```
mysterymath, line 97: A comment here seems like a good idea to indicate what function this performs.

mysterymath (approved):

examples/mega65/dma_audio.c
mlund: yep, good timing for embed to arrive in llvm-mos :-)

mos-platform/mega65/dma.hpp
mlund: fixed.

mlund: added.

mos-platform/mega65/dma.hpp
mlund: Added comment.

examples/mega65/drums.license
mlund, line 1: I got your first comment about this via email. Is the license finally OK?

mysterymath: Should be fine; I think I was being too tech brained for this; a music player really shouldn't be an "adaptation" of the music it plays.

PR #350, opened 2024-08-09
[cx16] Add C++ plasma example

mlund:
Commander X16 / VERA version of the C++ plasma example also found for C64 and MEGA65.

- [x] Tested on cx16emu, rom v47
- [x] Tested on hardware, rom v47
- [x] Passes pedantic compiler warning level


mysterymath (approved):
LGTM; thanks for the example!

Issue #334, opened 2024-06-02
Implement `atari` `fgetpos` and `fsetpos` using NOTE/POINT


mysterymath:
While the Atari DOS NOTE/POINT functionality isn't expressive enough to implement `fseek`, it is expressive enough to implement `fgetpos` and `fsetpos`, since these operate more-or-less the same way as NOTE and POINT, since the offsets they return are opaque.

pfusik:
I'd like to implement this.

pfusik:
How to handle `FILE` buffer?
`NOTE` will tell us sector/offset *after* the whole buffer we read.
Writing is easier: just flush the buffer, then `NOTE`.

I can only think of using `NOTE` at every entry of `fill_buffer`. The time overhead should be acceptable. Unfortunately, that grows all binaries that do not use `fgetpos`.
Then `fsetpos` does a `POINT` followed by skipping as many bytes as we had in the buffer. Possibly by just filling the buffer and moving the pointer.

mysterymath:
`NOTE`-ing on every `fill_buffer` SGTM; there's already quite a bit of stuff brought in just for standards compliance, and it shouldn't break the bank comparatively.

mysterymath:
Actually, I don't think that fixes anything. Say you call `fgetpos` with a full buffer, read a byte from the buffer, then call `fgetpos` again. The two values must be different, but there's no clear way to produce the sector and offset for the second one, particularly if the next byte came from a different sector.

I can't think of a way to implement this with buffered I/O without detailed introspection into the sector structure of files. But that would bake way too much into the libc.

pfusik:
I'm thinking of:

```c
typedef struct {
    unsigned short sector; // at the beginning of the buffer
    unsigned char byte_in_sector; // at the beginning of the buffer
    unsigned char byte_in_buffer; // assuming BUFSIZ=256
} fpos_t;
```

mysterymath:
> I'm thinking of:
> 
> ```c
> typedef struct {
>     unsigned short sector;          // at the beginning of the buffer
>     unsigned char offset_in_sector; // at the beginning of the buffer
>     size_t offset_in_buffer;        // size_t because of setvbuf
> } fpos_t;
> ```

Providing a custom `fpos_t` type per target may be... tricky, since `stdio-full.c` is currently built as part of common. But code also currently assumes that it's a file offset, and that's probably not tenable long-term. Some design and experimentation work is probably prudent around this.

[...]

PR #351, opened 2024-08-11
Fix leak on fopen error

pfusik:


pfusik:
Found this while working on #334
I'd like to follow-up with a refactor of the common endings `fopen` and `freopen`.

mysterymath (approved):
Nice catch; thanks! LGTM.

PR #352, opened 2024-08-11
Fix null dereference on tmpfile error

pfusik:


mysterymath (approved):
Subtle, but nice, thanks! LTGM.

PR #357, opened 2024-08-13
[Breaking] [cx16] SNES controller KERNAL function and example

mlund:
This adds an example of how to use the SNES controllers on Commander X16. This is a **breaking change** due to `cx16_k_joystick_get()` signature change. This is likely not a big issue as it has most likely not been used much in the wild, considering the incomplete bitmasks for buttons, see below.

- [x] Change signature of `cx16_k_joystick_get()` to return a three-byte struct, `JoyState` instead of a four-byte integer. This has the advantage of (1) being formally one byte shorter and (2) enables easy byte-wise bit masking (16 bit also exposed via union). The last byte is for checking if the joystick is attached and is cast to a bool which is cleaner than checking for -1. Finally this approach avoids duplication of Y.
- [x] Fix a bitmask bug `cx16.h` (Button B).
- [x] Add bitmask defines for missing buttons and change to explicitly typed enum.
- [x] Remove unused macros relevant only to cc65's general joystick support.
- [x] If on C++, enable inlined member functions. Illustrated in C++ example, testing all buttons.
- [x] `cx16_k_joystick_get()` now uses inline assembly for better optimization of register usage.
- [ ] ~~Added `common/include/stdbool.h` from clang to allow use of bools in C code.~~


mos-platform/cx16/cx16.h
```diff
@@ -433,10 +475,31 @@ void cx16_k_graph_set_font(void *fontaddr) __attribute__((leaf));
...
+
+/**
+ * Get joystick state using KERNAL routine `JOYSTICK_GET`.
+ *
+ * More information: https://github.com/X16Community/x16-docs
+ *
+ * @param joystick_num Keyboard joystick (0) or SNES controllers (1-4).
+ * @returns Struct with current status.
+ */
+inline struct JoyState cx16_k_joystick_get(const unsigned char joystick_num) {
```
mlund, line 487: Unsure if this function should be annotated with `__attribute__((leaf))`?

mos-platform/common/include/stdbool.h
mysterymath, line 1: Clang actually ships `stdbool.h` as part of it's compiler support headers, so we shouldn't need it in the SDK. Did you have difficulty using it in the SDK or something?

mos-platform/cx16/cx16.h
mysterymath: There's no need; `__attribute__((leaf))` has no effect on defined functions. The point is to describe the behavior of functions that the compiler can't see the implementation of.

mos-platform/cx16/cx16.h
```diff
@@ -118,29 +120,69 @@ extern "C" {
...
+      unsigned char data1; //!< Bits: A X FireL  FireR 1  1    1    1
+    };
+    unsigned short data; //!< data0 and data1 combined
+  };
+  bool detached; //!< True if joystick is disconnected
 
-#define JOY_FIRE2_MASK          JOY_BTN_2_MASK
-#define JOY_FIRE2(v)            ((v) & JOY_FIRE2_MASK)
+#ifdef __cplusplus
+  inline bool button_a() const { return !(data1 & JOY_BTN_A_MASK); }
```
mysterymath, line 160: `inline` is superfluous: all definitions directly in structs/classes are `inline`

mos-platform/cx16/cx16.h
```diff
@@ -433,10 +475,31 @@ void cx16_k_graph_set_font(void *fontaddr) __attribute__((leaf));
...
+ *
+ * More information: https://github.com/X16Community/x16-docs
+ *
+ * @param joystick_num Keyboard joystick (0) or SNES controllers (1-4).
+ * @returns Struct with current status.
+ */
+inline struct JoyState cx16_k_joystick_get(const unsigned char joystick_num) {
+  struct JoyState s;
+  __attribute__((leaf)) asm volatile(
+      "JOYSTICK_GET = $FF56\n"
```
mysterymath, line 490: There seems to be a disconnect between the symbol names here; only one has `__`.

mysterymath (commented):
This change should be tagged as [Breaking], since it changes the API.

mos-platform/common/include/stdbool.h
mlund: Hmm, I definitely had issues including the header, but now I cannot reproduce.
Reverted the change.

mos-platform/cx16/cx16.h
mlund: Removed inline keywords.

mos-platform/cx16/cx16.h
mlund: Good catch! `__JOYSTICK_GET` is defined one level up in `commodore/cbm_kernal.inc`. Removed local define.

mos-platform/cx16/cx16.h
```diff
@@ -433,10 +469,30 @@ void cx16_k_graph_set_font(void *fontaddr) __attribute__((leaf));
...
+
+/**
+ * Get joystick state using KERNAL routine `JOYSTICK_GET`.
+ *
+ * More information: https://github.com/X16Community/x16-docs
+ *
+ * @param joystick_num Keyboard joystick (0) or SNES controllers (1-4).
+ * @returns Struct with current status.
+ */
+inline struct JoyState cx16_k_joystick_get(const unsigned char joystick_num) {
```
mysterymath, line 481: nit: `const` isn't idiomadict for ABI here, since the argument is passed by value.

mlund: I like to keep `const` in function _definitions_ even when passing by value as a way of documenting the intent. Will remove it from the declaration when splitting as suggested below.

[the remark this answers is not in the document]
mlund: Ah, didn't know about the bitcode; nice. Will split the definition into a separate source file.

mysterymath (approved):
Thanks for the change; LGTM!

Issue #354, opened 2024-08-11
Double free in freopen on OOM


pfusik:
```c
static FILE *init_file(FILE *stream) {
  ...
  if ((rc->buffer = (char *)malloc(BUFSIZ)) == NULL) {
    /* No memory */
    free(rc);
    return NULL;
  }
```
```c
FILE *freopen(const char *restrict filename, const char *restrict mode,
              FILE *restrict stream) {
  ...
  if (init_file(stream) == NULL) {
    /* Re-init failed. */
    free(stream->filename);
    free(stream);
    return NULL;
  }
```

Wondering if `free(stream)` can attempt to free the memory just freed by `free(rc)` ?

Issue #356, opened 2024-08-11
Filename allocation problems in freopen


pfusik:
1. `filename = stream->filename;` leaks.
2. `free(stream->filename);` double free.
3. `strcpy(stream->filename, filename);` null-dereference on OOM (also in `fopen`).

pfusik:
[cppreference](https://en.cppreference.com/w/c/io/freopen) says:

> If filename is a null pointer, then the function attempts to reopen the file that is already associated with stream (it is implementation defined which mode changes are allowed in this case).
> Microsoft CRT version of freopen does not support any mode changes when filename is a null pointer and treats this as an error

If that's good enough for Microsoft CRT, I suggest to make it simple and only store the filename for `DELONCLOSE`/`tmpfile`.
At least on the Atari, you can delete a file by its name, but not a handle.

PR #361, opened 2024-08-18
Improve freopen

pfusik:


pfusik:
Fixes #354 and #356

mos-platform/common/c/stdio-full.c
```diff
@@ -128,36 +128,7 @@ static unsigned filemode(const char *const mode) {
...
-  rc->status = FREEBUFFER;
-  rc->del_filename = NULL;
-
-  // TODO: Setting mbstate
-
-  return rc;
-}
-
-static signed char stdio_open(const char *const filename, unsigned int mode) {
+static FILE *stdio_open(FILE *stream, const char *filename, unsigned int mode) {
```
mysterymath, line 211: This needs a comment describing the relationship between the argument and the return value. In particular, it wasn't clear that this takes ownership of the stream argument, and that that argument must be dynamically allocated.

mos-platform/common/c/stdio-full.c
```diff
@@ -292,8 +281,13 @@ int fclose(FILE *stream) {
   /* Free stream */
   if (stream != stdin && stream != stdout && stream != stderr)
     free(stream);
+}
 
-  return 0;
+int fclose(FILE *stream) {
+  int rc = stdio_close(stream);
+
```
mysterymath, line 288: nit: delete line.

mos-platform/common/c/stdio-full.c
```diff
@@ -87,6 +85,86 @@ void _stdio_closeall(void) {
...
+    if (remove(stream->del_filename) != 0)
+      rc = EOF;
+    free(stream->del_filename);
+  }
+  return rc;
+}
+
+static void free_stream(FILE *stream) {
+  /* Remove stream from list */
+  if (filelist == stream)
```
mysterymath, line 148: nit: One aspect of the LLVM style I'd like to keep is that if one clause of an if/else has braces; the other must also. Making them the two clauses visually similar tends to help with finding their ranges at a glance.

mos-platform/common/c/stdio-full.c
```diff
@@ -367,110 +335,20 @@ FILE *freopen(const char *restrict filename, const char *restrict mode,
...
-
-  if (filename == NULL) {
-    /* Attempt to change mode without closing stream */
-    switch (change_mode(stream, fmode)) {
-    case INT_MIN:
-      /* fail completely */
-      return NULL;
-
-    case 0:
-      /* fail; try close / reopen */
```
mysterymath, line 382: This PR should have `[Breaking]` in its description, since this is a change to existing implementation-defined behavior.

mysterymath (commented):
This PR description needs an itemized summary of the contained changes; otherwise it will be difficult to write the release notes post-squash. Particularly since it's a breaking change.

mos-platform/common/c/stdio-full.c
pfusik: Done.
`stream` is not necessarily `malloc`ed - it can be `stdout` from `freopen`. The old code was doing `free` without checking that - this is one of the bugs I'm fixing.

mos-platform/common/c/stdio-full.c
pfusik: Done.

mos-platform/common/c/stdio-full.c
pfusik: Done.

pfusik:
Do we need to squash this? These are three related but separate changes. I would prefer to squash only the fixup commit that addresses review comments.

mysterymath:
> Do we need to squash this? These are three related but separate changes. I would prefer to squash only the fixup commit that addresses review comments.

I suppose we can do that.

mysterymath (approved):
LGTM, thanks!
I'll hold off on committing until the squash; let me know.

mos-platform/common/c/stdio-full.c
```diff
@@ -208,6 +208,10 @@ static unsigned filemode(const char *const mode) {
   return rc;
 }
 
+// Initializes stream by opening the handle
+// and initializing most of structure members.
+// stream->buffer, stream->bufsize and stream-next must be initialized on entry.
```
pfusik, line 213: Fixing `stream-next` typo while squashing.

mos-platform/common/c/stdio-full.c
mysterymath: Since we're rebasing; this should be done on the individual commit that introduces the breaking change.

pfusik: Please read the new commit messages. If we split the first commit, the motivation could be become unclear.

PR #362, opened 2024-08-21
Fix `fopen(filename, "a")` misinterpreted as `fopen(filename, "w")`

### LINKED ISSUE #353: fopen(filename, "a") truncates?
Looking at the code:

```c
static unsigned filemode(const char *const mode) {
  unsigned rc = 0;

  switch (mode[0]) {
  case 'r':
    rc |= FREAD;
    break;

  case 'w':
    rc |= FWRITE;
    break;

  case 'a':
    rc |= FAPPEND | FWRITE;
    break;
  }
```

```c
static signed char stdio_open(const char *const filename, unsigned int mode) {
  int osmode;

  if (mode & FRW)
    osmode = O_RDWR;
  else if (mode & FREAD)
    osmode = O_RDONLY;
  else
    osmode = O_WRONLY;

  if (mode & (FWRITE | FAPPEND))
    osmode |= O_CREAT;

  if (mode & FWRITE)
    osmode |= mode & FEXCL ? O_EXCL : O_TRUNC;
  else if (mode & FAPPEND)
    osmode |= O_APPEND;
```

`"a"` will set `O_TRUNC` i.e. truncate an existing file.
I think we should either not set `FWRITE` or reorganize the logic in `stdio_open`.

pfusik:
Fix #353

### DIFF
--- a/mos-platform/common/c/stdio-full.c
+++ b/mos-platform/common/c/stdio-full.c
@@ -228,12 +228,15 @@ static FILE *stdio_open(FILE *stream, const char *filename, unsigned int mode) {
     osmode = O_WRONLY;
 
   if (mode & (FWRITE | FAPPEND))
-    osmode |= O_CREAT;
+    osmode |= O_CREAT;    // file doesn't exist: create
 
-  if (mode & FWRITE)
-    osmode |= mode & FEXCL ? O_EXCL : O_TRUNC;
-  else if (mode & FAPPEND)
-    osmode |= O_APPEND;
+  if (mode & FAPPEND) {
+    osmode |= O_APPEND;   // file exists: append
+  } else if (mode & FWRITE) {
+    osmode |= mode & FEXCL
+               ? O_EXCL   // file exists: fail
+               : O_TRUNC; // file exists: truncate
+  }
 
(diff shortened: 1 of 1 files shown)


mos-platform/common/c/stdio-full.c
```diff
@@ -169,12 +169,15 @@ static signed char stdio_open(const char *const filename, unsigned int mode) {
...
 
   if (mode & (FWRITE | FAPPEND))
-    osmode |= O_CREAT;
+    osmode |= O_CREAT;    // file doesn't exist: create
 
-  if (mode & FWRITE)
-    osmode |= mode & FEXCL ? O_EXCL : O_TRUNC;
-  else if (mode & FAPPEND)
-    osmode |= O_APPEND;
+  if (mode & FAPPEND)
```
mysterymath, line 174: nit: braces around `if` clause if `else` clause has them

mysterymath (approved):
LGTM; thanks!

mos-platform/common/c/stdio-full.c
pfusik: Done

PR #364, opened 2024-08-24
Fix use after free in `_stdio_closeall`

pfusik:
While this was probably harmless because we are single-threaded, the allocator could generally store its internal structure in the freed block.


mysterymath (approved):
Ah nice! This isn't harmless, because our `malloc` absolutely does do that.

PR #369, opened 2024-08-26
[cx16] Add missing asm sections and progbits

mlund:


mlund:
I just read the wiki and realise that `R` means _retain_ instead of not read-only. I guess we don't want that(?)

mysterymath:
> I just read the wiki and realise that `R` means _retain_ and not _read-only_. I guess we don't want that(?)

That's right, the only things that should be marked `R` are those sections where it would be incorrect for the linker to *ever* remove them, even if completely unused.

mysterymath (approved):

PR #370, opened 2024-08-26
[cx16] Add VERA PSG registers

mlund:


mos-platform/cx16/cx16.h
```diff
@@ -364,6 +368,22 @@ struct __ym2151 {
...
+            unsigned char freq_lo; //!< Frequency word (7:0)
+            unsigned char freq_hi; //!< Frequency word (15:8)
+        };
+    };
+    unsigned char volume;   //!< Left (bit 7); right (bit 6); volume (bit 5:0)
+    unsigned char waveform; //!< Waveform (bit 7:6) and pulse width (bit 5:0)
+};
+#if (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L) || (defined(__cplusplus) && __cplusplus >= 201103L)
+static_assert(sizeof(struct __vera_psg) == 4, "struct __vera_psg must be 4 bytes");
+#endif
```
mysterymath, line 385: Doesn't there need to be a `#define` that corresponds to this struct definition?

mlund: The PSG registers live at a 17-bit VERA video address and data must first be prepared in low ram; then copied to/from VERA. The naming however can be discussed: I stuck with the CC65 inspired `__` prefix which hints it's private, and the convention is used for all other hardware registers. As we in this case cannot make a pointer `#define`, perhaps it's better to call it `VeraPSG`(?)

mysterymath (approved):

PR #375, opened 2024-09-17
created geos-cbm

Dirk-Schnorpfeil:
Hi all,

i added support for GEOS subsystem on commodore machines. Also added an example (SamVlir) in example directory, which is a port from assembler to C of the SamVlir demo application of GeoProgrammer.


mos-platform/geos-cbm/Makefile
mlund, line 1: This file is generated by CMake and should not be part of the repository.

examples/geos-cbm/CMakeLists.txt
mlund, line 1: This seems to start a new CMake project in an example sub-folder. If possible, the `geos-cbm` should be included with `add_subdirectory()` inside `examples/CMakeLists.txt`

mos-platform/geos-cbm/Makefile
Dirk-Schnorpfeil: Right, will remove it.

examples/geos-cbm/CMakeLists.txt
Dirk-Schnorpfeil: It is included per add_subdirectory. But wouldnt it need its only CMakeList to compile a program which consists of multiple source files?

mlund:
I think you probably need to run `clang-format -i` on the .h and .c files. This will pick up the formatting style of the SDK, found in root level (`.clang-format`).

mos-platform/geos-cbm/link.ld
mlund, line 1: Is an empty file needed?

mos-platform/geos-cbm/geos_crt.c
```diff
@@ -0,0 +1,2013 @@
...
+/***********************************************************************
+ * GEOS Graphic routines
+ ***********************************************************************/
+
+void DrawPoint(uint16_t x, uint8_t y, bool recover, bool background)
+{
+  __r3 = x;
+  __r11L = y;
+  if (recover) {
+    asm volatile("lda #0xff \n");
```
mlund, line 23: Not entirely sure if it's needed, but the wiki suggest that for performance reasons, inline asm should be marked with `__attribute__((leaf))`. See https://llvm-mos.org/wiki/Optimization_guide.

mos-platform/geos-cbm/geos_crt.c
```diff
@@ -0,0 +1,2013 @@
...
+
+/***********************************************************************
+ * GEOS Character routines
+ ***********************************************************************/
+
+void PutString(uint16_t x, uint8_t y, const char *str)
+{
+  __r11 = x;
+  __r1H = y;
+  __r0 = (uint16_t)str;
```
mlund, line 216: Is it OK to overwrite `__r0`, or should it be added to the clobber list? I do not fully understand the custom register setup, so it may be perfectly fine.

mos-platform/geos-cbm/geos_crt.c
```diff
@@ -0,0 +1,2013 @@
...
+}
+
+void CopyFString(uint8_t *src, uint8_t *dest, uint8_t count)
+{
+  __r0 = (uint16_t)src;
+  __r1 = (uint16_t)dest;
+  asm volatile(
+      "ldx   #2 \n"
+      "ldy   #4 \n"
+      "jsr   __CopyFString " : "=a" (count) : "a" (count) : "x","y","c","v");
```
mlund, line 558: No need to specify an output operand as `count` is passed by value and never leaves the function.

mos-platform/geos-cbm/geos_crt.c
```diff
@@ -0,0 +1,2013 @@
...
+}
+
+int8_t CmpString(char *src, char *dest)
+{
+  uint8_t result;
+
+  __r0 = (uint16_t)src;
+  __r1 = (uint16_t)dest;
+  asm volatile(
```
mlund, line 567: Just a hint that multi-line inline asm may be nicely written using raw strings, see here for an example: https://godbolt.org/z/hG57Eba4v

mos-platform/geos-cbm/clang.cfg
```diff
@@ -0,0 +1,8 @@
+-Os
+-flto
+-fnonreentrant
+-mllvm -inline-threshold=0
```
mlund, line 4: Is there a particular reason for setting `inline-threshold=0`?

mos-platform/geos-cbm/clang.cfg
```diff
@@ -0,0 +1,8 @@
+-Os
+-flto
+-fnonreentrant
```
mlund, line 3: OK, I see you have enabled nonreentrant whereby you may ignore my previous comment on `__attribute__((leaf))` on inline asm.

mlund:
Should this target be placed under `mos-hardware/commodore`? This would enable access to common CBM jump tables for ROM functions.

mos-platform/geos-cbm/link.ld
Dirk-Schnorpfeil: Without that I get compilation/link errors

mos-platform/geos-cbm/clang.cfg
Dirk-Schnorpfeil: It was inlining way to much. Also Code from overlay Module. So I tried that.

mos-platform/geos-cbm/clang.cfg
mysterymath: This really isn't an okay thing to have by default in a `clang.cfg` file. It really dramatically  changes the semantics of C globally, and users should have to opt in themselves.

mos-platform/geos-cbm/clang.cfg
mlund: I prefer to use `-Os -flto` and let the compiler decide what to inline. Or did that produce (too) large binaries?

examples/geos-cbm/CMakeLists.txt
Dirk-Schnorpfeil: OK, but then i need to write the add_executable into the CMakeLists.txt in example directory?

mos-platform/geos-cbm/geos_constants.h
```diff
@@ -0,0 +1,710 @@
...
+ ***********************************************************************/
+
+#ifndef GEOS_CONSTANTS_H
+#define GEOS_CONSTANTS_H
+
+/***********************************************************************
+ * Miscellaneous defines
+ ***********************************************************************/
+
+typedef unsigned char uint8_t;              /* an unsigned byte */
```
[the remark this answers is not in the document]
Dirk-Schnorpfeil: Replaced by include of stdbool. and stdint.h

mos-platform/geos-cbm/geos_crt.c
Dirk-Schnorpfeil: Replaced all occurences

mos-platform/geos-cbm/geos_crt.c
Dirk-Schnorpfeil: Problem is that a is overwritten in __CopyFString. When i put a in input and clobber list i get an error message. Best way i found to tell compiler that a is overwritten was by putting it in the output list.

Any idea is welcome :-)

mos-platform/geos-cbm/geos_crt.c
Dirk-Schnorpfeil: Next time i know :-) I was searching for it but did not find it.

mos-platform/geos-cbm/clang.cfg
Dirk-Schnorpfeil: Removed it from clang.cfg file.

examples/geos-cbm/SamVlir.txt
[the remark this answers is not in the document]
Dirk-Schnorpfeil: It was for my convenience checking the produced code. Removed it.

Dirk-Schnorpfeil:
> Should this target be placed under `mos-hardware/commodore`? This would enable access to common CBM jump tables for ROM functions.

I feel it should not because using the jump tables would crash the program. A lot of zero-page locations which are used by ROM functions are now used by the compiler. All GEOS calls which might call ROM functions first swap the zero page locations which are used back in and after the call again out.

Dirk-Schnorpfeil:
> I think you probably need to run `clang-format -i` on the .h and .c files. This will pick up the formatting style of the SDK, found in root level (`.clang-format`).

I tried, but after that the code looked more ugly. The header files are formatted to look close to the symbol definitions in the geoProgrammer package...

Dirk-Schnorpfeil:
Added a README.md to .../mos-platform/geos-cbm

mlund:
> > I think you probably need to run `clang-format -i` on the .h and .c files. This will pick up the formatting style of the SDK, found in root level (`.clang-format`).
> 
> I tried, but after that the code looked more ugly. The header files are formatted to look close to the symbol definitions in the geoProgrammer package...

Sections can be omitted from formatting by clamping in between `// clang-format off` and `// clang-format on` comments. I'd say that this should be done only for exceptional cases where formatting for some reason obstructs readability.

mos-platform/geos-cbm/geos_crt.c
Dirk-Schnorpfeil: Because I rember seeing generated code where the later code was dependend on the value in a. Thats why i would not like to remove it.

mos-platform/geos-cbm/geos_crt.c
```diff
@@ -0,0 +1,2267 @@
...
+  __r11H = right_skip;
+  __r12 = top_skip;
+  __r13 = (uint16_t)app_input;
+  __r14 = (uint16_t)sync;
+
+  __attribute__((leaf))
+  asm volatile("jsr __BitOtherClip " : : : "a","x","y","c","v");
+}
+
+__attribute__((leaf))
```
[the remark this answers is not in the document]
Dirk-Schnorpfeil: But it doesn´t harm either, right? So i would keep it as a little service to the compiler :-)

mos-platform/geos-cbm/geos_crt.c
Dirk-Schnorpfeil: I could not reproduce it so i remove the output operands where not needed.

mos-platform/geos-cbm/geos_crt.c
```diff
@@ -0,0 +1,2267 @@
...
+void DrawLine(uint16_t left, uint8_t top, uint16_t right, uint8_t bottom, bool recover, bool background)
+{
+  __r3 = left;
+  __r11L = top;
+  __r4 = right;
+  __r11H = bottom;
+
+  if (recover) {
+    __attribute__((leaf))
+    asm volatile("lda #0xff \n");
```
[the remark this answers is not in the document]
Dirk-Schnorpfeil: Fixed it.

mos-platform/geos-cbm/geos_crt.c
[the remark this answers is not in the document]
Dirk-Schnorpfeil: Fixed it.

Dirk-Schnorpfeil:
Hi, one question: how does this go on from here? I assume you will merge the changes to your main branch. But what is the process when I add more functionality or fix bugs?

Would I get the right to make changes to the gros-cbm subdirs or would we go through the pull process every time?

Thanks a lot,

Dirk

mlund:
> Hi, one question: how does this go on from here? I assume you will merge the changes to your main branch. But what is the process when I add more functionality or fix bugs?
> 
> Would I get the right to make changes to the gros-cbm subdirs or would we go through the pull process every time?
> 
> Thanks a lot,
> 
> Dirk

@mysterymath is the one ultimately doing the merging. All following changes go through a similar process, but PR's are then usually small and incremental, and less extensive as this first commit.

mlund:
> I feel it should not because using the jump tables would crash the program. A lot of zero-page locations which are used by ROM functions are now used by the compiler. All GEOS calls which might call ROM functions first swap the zero page locations which are used back in and after the call again out.

I'm not too familiar with geos, but know of C64, C128 and an Apple II version. There's also work in progress for MEGA65. How much of the current PR is specific to Commodore? Could some headers or code be shared?

mos-platform/geos-cbm/geos_constants.h
```diff
@@ -0,0 +1,757 @@
...
+#define GEOS_CONSTANTS_H
+
+/***********************************************************************
+ * Miscellaneous defines
+ ***********************************************************************/
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#if 0
```
mlund, line 21: This section is not needed.

mos-platform/geos-cbm/geos_constants.h
```diff
@@ -0,0 +1,757 @@
...
+#define AUX(x) #x
+#define STRINGIFY(x) AUX(x)
+
+/***********************************************************************
+ * Hardware related defines
+ ***********************************************************************/
+
+/* values for CPU_DATA - C64 memory config */
+
+#define RAM_64K 0x30        /* 64K RAM */
```
mlund, line 46: If instead using `/** ... */` (double `**`) for comments, Doxygen will pick it up.

mos-platform/geos-cbm/geos_routines.h
```diff
@@ -0,0 +1,311 @@
...
+ * Jump addresses within disk drivers: these are only valid for non-1541
+ * disk drive types, and for the 128 versions of the 1541 driver.
+ ***********************************************************************/
+disk_err_t Get1stDirEntry(dir_entry_t **dir_entry);
+disk_err_t GetNxtDirEntry(dir_entry_t **dir_entry, bool *end_of_dir);
+disk_err_t GetOffPageTrSc(bool *no_geos_disk, tr_se_pair_t *border_block);
+disk_err_t AllocateBlock(tr_se_pair_t block_addr);
+disk_err_t ReadLink(tr_se_pair_t block_addr, uint8_t *buffer);
+
+/* undocumented Official and Hitchhikers Guides
```
mlund, line 34: Remove

mos-platform/geos-cbm/geos_crt.c
```diff
@@ -0,0 +1,1910 @@
...
+}
+
+void DrawLine(uint16_t left, uint8_t top, uint16_t right, uint8_t bottom,
+              bool recover, bool background) {
+  __r3 = left;
+  __r11L = top;
+  __r4 = right;
+  __r11H = bottom;
+
+  if (recover) {
```
mlund, line 102: same comment as for DrawPoint

mos-platform/geos-cbm/geos_constants.h
Dirk-Schnorpfeil: I am not sure what type uint8_t now is, when including stdint.h. One nice effect when defining it myself as unsigned char was that one could give a char* instead of an uint8_t*, therefore saving casts in the source code...

For now i will remove the #if 0 block, but maybe later I need to refine uint8_t to unsigned char again.

Dirk-Schnorpfeil:
> > I feel it should not because using the jump tables would crash the program. A lot of zero-page locations which are used by ROM functions are now used by the compiler. All GEOS calls which might call ROM functions first swap the zero page locations which are used back in and after the call again out.
> 
> I'm not too familiar with geos, but know of C64, C128 and an Apple II version. There's also work in progress for MEGA65. How much of the current PR is specific to Commodore? Could some headers or code be shared?

I am also not familiar with GEOS for apple, atari and mega65. At cc65 these are seperate targets on the same level.

mos-platform/geos-cbm/geos_constants.h
mlund: `uint8_t` and `unsigned char` can be used interchangeably: https://godbolt.org/z/MEa3cezGY

mos-platform/geos-cbm/geos_crt.c
```diff
@@ -0,0 +1,1910 @@
...
+
+/***********************************************************************
+ * GEOS Graphic routines
+ ***********************************************************************/
+
+void DrawPoint(uint16_t x, uint8_t y, bool recover, bool background) {
+  __r3 = x;
+  __r11L = y;
+  if (recover) {
+    if (background) {
```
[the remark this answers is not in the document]
Dirk-Schnorpfeil: Much better code!! Changed it.

mos-platform/geos-cbm/vlir.ld
mysterymath, line 1: Why not just call this `link.ld`? That's the convention for the main linker script for each platform, and that appears to be the case for this one.

mysterymath (commented):
Sorry for taking so long; I've had a lot in my queue before this PR. 

Here's a few drive-by comments; there's a *lot* to churn through here, so I'll be looking through it in drips and drabs as I get time.

mos-platform/geos-cbm/vlir.ld
Dirk-Schnorpfeil: There are 3 different types of GEOS application: VLlR (overlayed), SEQ (non overlayed) and Desk Accessories. 

Each of them will need their own ld file. I want to supply seq.ld and acc.ld later.

mos-platform/geos-cbm/geos.h
```diff
@@ -0,0 +1,23 @@
...
+#ifndef GEOS_H
+#define GEOS_H
+
+#include "geos_constants.h"
+#include "geos_memorymap.h"
+#include "geos_types.h"
+#include "geos_symbols.h"
+#include "geos_routines.h"
+
+#define noinline __attribute__((noinline))
```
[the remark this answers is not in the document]
Dirk-Schnorpfeil: Well, one need to know how to handle the harsh mistress :-)

I like that abbreviated style but will change it as you suggested.

mos-platform/geos-cbm/vlir.ld
mysterymath: Would it be feasible to make these sub-targets? Maybe geos-cbm-vlir, geos-cbm-seq, and geos-cbm-acc?

mysterymath:
Another high level procedural question to @Dirk-Schnorpfeil : Are you intending to stick around and be somewhat available to provide support and answer questions about this, e.g. on our Discord?

Most targets are simple enough that I can fill that role myself, but this one has a ton of very specific GEOS knowledge baked in, and I'm out of my depth. I can advise on all the llvm-mos-isms, but I wouldn't be able to answer any questions about the specifics of how it interacts with GEOS, or maybe even the overlay model.

It's not necessarily make or break either way, but I might not end up being able to keep this working long term as we pull in changes from upstream LLVM and things subtly change. If users discover breakages, I'd either need to point them your way or advertise that the platform is increasing degrees of broken. If it gets too broken, I'd eventually need to pull it out.

This is a pretty long term concern though; it doesn't affect all that much in the here and now. Just something to think about.

mos-platform/geos-cbm/README.md
```diff
@@ -0,0 +1,415 @@
...
+You can change the size of the overlay memory to e.g. 2k bytes by adding the following option to the compile command line:
+
+    -Xlinker --symdef=__overlaysize_param=0x2000
+
+## Importing image data
+
+For e.g. icons in GEOS you need image data in a compressed format supported by GEOS. You need to supply it as an byte array like this:
+
+    uint8_t Icon1Picture[] = {
+	    /*
```
mysterymath, line 362: Might be worth mentioning `#embed` now that clang has support for it.

mos-platform/geos-cbm/README.md
```diff
@@ -0,0 +1,415 @@
...
+	    if (req_mod_no != curModule) {
+		    curModule = req_mod_no;
+		    ReadFile(OVERLAYADDR, OVERLAYSIZE, swapTable[req_mod_no-1]);
+	    }
+    }
+
+`OVERLAYADDR` and `OVERLAYSIZE` are defined by the linker and default to `0x5000` and `0x1000` (for details see the `vlir.ld` linker script in `.../mos-platform/geos-cbm/vlir.ld`).
+You can change the size of the overlay memory to e.g. 2k bytes by adding the following option to the compile command line:
+
+    -Xlinker --symdef=__overlaysize_param=0x2000
```
mysterymath, line 355: The NES targets provide preprocessor macros for setting symbols like this, which can be easier to use than piping a command line flag through whatever build system. Both are reasonable approaches, but one of the real advantages using linker symbols for configuration is that the values can live right in program code.

See https://github.com/llvm-mos/llvm-mos-sdk/blob/main/mos-platform/nes/ines.h for examples.

mos-platform/geos-cbm/vlir.ld
Dirk-Schnorpfeil: Passing the linker file via -T was the future plan. Subtargets are possible, but i would leave that for a future update. I just followed the convention used in cc65...

I will check if it is possible to combine all different geos target types in one vlir file and drive the decisions with setting some symbols.

Lets first find out if someone uses this target. Then i will be happy to supply more...

Dirk-Schnorpfeil:
> Another high level procedural question to @Dirk-Schnorpfeil : Are you intending to stick around and be somewhat available to provide support and answer questions about this, e.g. on our Discord?
> 
> Most targets are simple enough that I can fill that role myself, but this one has a ton of very specific GEOS knowledge baked in, and I'm out of my depth. I can advise on all the llvm-mos-isms, but I wouldn't be able to answer any questions about the specifics of how it interacts with GEOS, or maybe even the overlay model.
> 
> It's not necessarily make or break either way, but I might not end up being able to keep this working long term as we pull in changes from upstream LLVM and things subtly change. If users discover breakages, I'd either need to point them your way or advertise that the platform is increasing degrees of broken. If it gets too broken, I'd eventually need to pull it out.
> 
> This is a pretty long term concern though; it doesn't affect all that much in the here and now. Just something to think about.

Hi, yes i would be around and do  maintenance for that geos-cbm target. In a range that can be done having a full time job and family of cause  :-)

mos-platform/geos-cbm/README.md
Dirk-Schnorpfeil: Just to understand it better: Defining a symbol with .globl would have the same effect as defining it per --symdef?

If so, that would be a much simpler solution.

Can you please confirm?

mos-platform/geos-cbm/README.md
Dirk-Schnorpfeil: Embed would not do it even if one copies a Photo Scrap file. First the width and height bytes have to be removed and only the remaining bytes should be used as the compacted image data.

mysterymath:
> > Another high level procedural question to @Dirk-Schnorpfeil : Are you intending to stick around and be somewhat available to provide support and answer questions about this, e.g. on our Discord?
> > Most targets are simple enough that I can fill that role myself, but this one has a ton of very specific GEOS knowledge baked in, and I'm out of my depth. I can advise on all the llvm-mos-isms, but I wouldn't be able to answer any questions about the specifics of how it interacts with GEOS, or maybe even the overlay model.
> > It's not necessarily make or break either way, but I might not end up being able to keep this working long term as we pull in changes from upstream LLVM and things subtly change. If users discover breakages, I'd either need to point them your way or advertise that the platform is increasing degrees of broken. If it gets too broken, I'd eventually need to pull it out.
> > This is a pretty long term concern though; it doesn't affect all that much in the here and now. Just something to think about.
> 
> Hi, yes i would be around and do maintenance for that geos-cbm target. In a range that can be done having a full time job and family of cause :-)

That's good to hear, and I'm in a similar boat. This is a hobby thing all around; I'm mostly just trying to do what I can to preserve the knowledge of how things work. :)

mos-platform/geos-cbm/README.md
mysterymath: Yep, that's correct.

mos-platform/geos-cbm/geos.h
```diff
@@ -0,0 +1,20 @@
...
+ * with llvm-mos C compiler.
+ *
+ * Transcribed from geoProgrammer manual by DiSc.
+ *
+ ***********************************************************************/
+
+#ifndef GEOS_H
+#define GEOS_H
+
+#include "geos_constants.h"
```
mysterymath, line 14: This should begin with the usual C++ wrappers to make sure that the symbols defined within use the C API in C++.
See most of the other C headers in the SDK for how to do this.

mos-platform/geos-cbm/geos_constants.h
```diff
@@ -0,0 +1,752 @@
...
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#define LOW_BYTE(x) (*((unsigned char *)&(##x) + 0))
+#define HIGH_BYTE(x) (*((unsigned char *)&(##x) + 1))
+
+#define WORD(w) (w & 0xff), (w >> 8)
+
+#define AUX(x) #x
```
mysterymath, line 26: Internal defines like this should be `#undef`ed after they are used.

mos-platform/geos-cbm/geos_constants.h
```diff
@@ -0,0 +1,752 @@
...
+#define GEOS_CONSTANTS_H
+
+/***********************************************************************
+ * Miscellaneous defines
+ ***********************************************************************/
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#define LOW_BYTE(x) (*((unsigned char *)&(##x) + 0))
```
mysterymath, line 21: This could use a comment describing what's legal to pass to it; it doesn't look likely that these will work on arbitrary C expressions.

mos-platform/geos-cbm/geos_constants.h
```diff
@@ -0,0 +1,752 @@
...
+#define CKRNLBASIOIN 0x40 /** kernal, I/O and basic ROM's mapped into memory   \
+                           */
+#define CKRNLIOIN 0x4E    /** Kernal ROM and I/O space mapped in */
+
+/***********************************************************************
+ * Menu Equates
+ ***********************************************************************/
+
+/** Menu types */
+typedef enum __attribute__((__packed__)) {
```
mysterymath, line 53: We default to packed enums in llvm-mos; this should be superfluous.

mos-platform/geos-cbm/geos_constants.h
```diff
@@ -0,0 +1,752 @@
...
+  HORIZONTAL = 0b00000000,
+  VERTICAL = 0b10000000,
+  CONSTRAINED = 0b01000000,
+  UN_CONSTRAINED = 0b00000000
```
mysterymath, line 57: `UNCONSTRAINED`? It's an English word, so it's odd to see it in snake_case. Disregard if this was taken verbatim from a more authoritative source.

mos-platform/geos-cbm/geos_constants.h
```diff
@@ -0,0 +1,752 @@
...
+#define BOLD_BIT 6
+#define REVERSE_BIT 5
+#define ITALIC_BIT 4
+#define OUTLINE_BIT 3
+#define SUPERSCRIPT_BIT 2
+#define SUBSCRIPT_BIT 1
+
+/** Putchar constants */
+
+#define EOF 0       /** end of text object */
```
mysterymath, line 117: `EOF` and `NULL` are defined elsewhere in the C standard library; do they need to be defined here?
It's also the case that many of the below have standard ASCII character escapes (e.g., BS is `\b`). Do we need all these?

Again, if this is intending to be backwards compatbile with something else, that's another question. NULL and EOF may still cause issues; in particular, NULL is usually pointer typed, and EOF is usually negative one.

mos-platform/geos-cbm/geos_constants.h
```diff
@@ -0,0 +1,752 @@
...
+#define OUTLINEON 26    /** turn on OUTLINE characters */
+#define PLAINTEXT 27    /** plain text mode */
+#define USELAST 127     /** erase character */
+#define SHORTCUT 128    /** shortcut character */
+
+/***********************************************************************
+ * Keyboard Equates
+ ***********************************************************************/
+
+#define KEY_INVALID = 31
```
mysterymath, line 148: The equal signs here are almost certainly an error.

mos-platform/geos-cbm/geos_memorymap.h
```diff
@@ -0,0 +1,34 @@
...
+#define EXP_BASE 0xdf00      /** base address of ram expansion unit */
+#define MOUSE_BASE_128 0xfd00
+#define MOUSE_JMP_128 0xfd00
+#define MOUSE_JMP 0xfe80  /** start of mouse jump table */
+#define MOUSE_BASE 0xfe80 /** start of input driver */
+#define END_MOUSE_128 0xfe80
+#define config 0xff00
+#define END_MOUSE 0xfffa /** end of input driver */
+
+#endif
```
mysterymath, line 34: nit: End all files in newlines. (Github shows files without using this symbol).

mos-platform/geos-cbm/geos_memorymap.h
```diff
@@ -0,0 +1,34 @@
...
+#define SCREEN_BASE 0xa000   /** base of foreground screen */
+#define OS_ROM 0xc000        /** start of OS code space */
+#define OS_JUMPTAB 0xc100    /** start of GEOS jump table */
+#define EXP_BASE 0xdf00      /** base address of ram expansion unit */
+#define MOUSE_BASE_128 0xfd00
+#define MOUSE_JMP_128 0xfd00
+#define MOUSE_JMP 0xfe80  /** start of mouse jump table */
+#define MOUSE_BASE 0xfe80 /** start of input driver */
+#define END_MOUSE_128 0xfe80
+#define config 0xff00
```
mysterymath, line 31: Is there a particular reason the case differs for this definition?

mysterymath:
Oh, and before I forget, @Dirk-Schnorpfeil, thanks for putting this together! Having llvm-mos able to target GEOS is EXTREMELY NEAT, particularly with all of the weird overlay trimmings!

I'll keep churning through these files over the upcoming couple days; looking forward to getting this polished up!

mos-platform/geos-cbm/geos_constants.h
Dirk-Schnorpfeil: These definitions are from the GEOS original documentation. As no one should include stdio etc I would leave the definitions in here.

mos-platform/geos-cbm/geos_constants.h
Dirk-Schnorpfeil: Its written that way in the GEOS original documentation.

mos-platform/geos-cbm/geos_constants.h
Dirk-Schnorpfeil: Hmm, maybe i do not get you. But if I #undef AUX the Macro using it will fail at the place where it is used.

mos-platform/geos-cbm/geos_constants.h
Dirk-Schnorpfeil: Not needed anymore. I removed it.

mos-platform/geos-cbm/geos_constants.h
Dirk-Schnorpfeil: That does not seem to hold true. I put some Debug_Asserts on the needed size of structures and they fail if I remove the attribute from the enums.

mos-platform/geos-cbm/geos.h
```diff
@@ -0,0 +1,18 @@
...
+ * geos.h
+ *
+ * This file contains definitions for use in GEOS applications
+ * with llvm-mos C compiler.
+ *
+ * Transcribed from geoProgrammer manual by DiSc.
+ *
+ ***********************************************************************/
+
+#pragma once
```
mysterymath, line 11: We prefer specific include guards in the SDK for subtle semantic reasons (mostly to do with the complexity of "flie identity").

mos-platform/geos-cbm/geos_constants.h
mysterymath: Nope. you're right. Sorry.

mos-platform/geos-cbm/geos_constants.h
mysterymath: I think EOF is probably fine, but NULL is also defined in `stdlib.h` and `string.h`, and both are likely to be useful on GEOS (right?). As-is, one will get a strange duplicate warning error for the two definitions, and a failure on `-Werror`.

Presumably the GEOS documents were for an assembly language interface, not C, right? The canonical way to spell a null character in C is `\0`, and that does also appear to be the correct character here. C NULL is also explicitly intended to be used as null pointer constant, not a null character constant.

This is a place where some negotiation needs to happen between C standard practice and legacy practices. If we do need these constants, special casing them to GEOS_EOF and GEOS_NULL seems warranted.

mos-platform/geos-cbm/geos_constants.h
mysterymath: I set up a little example with this in godbolt, and it looks to be emitting a single-byte enum. Maybe something else is going on?

I'll have to try this on my local install when I get home; it's also possible that godbolt is out of date and that this is a recent regression. We do consider the effects of `-fshort-enums` to be part of the llvm-mos ABI, and that has the effect of placing packed on every enum definition.

See https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:2,endLineNumber:12,positionColumn:2,positionLineNumber:12,selectionStartColumn:2,selectionStartLineNumber:12,startColumn:2,startLineNumber:12),source:'%23include+%3Cstdio.h%3E%0A%0Avolatile+unsigned+s%3B%0Avoid+f(void)+%7B%0A++++enum+%7B%0A++++++HORIZONTAL+%3D+0b00000000,%0A++++++VERTICAL+%3D+0b10000000,%0A++++++CONSTRAINED+%3D+0b01000000,%0A++++++UN_CONSTRAINED+%3D+0b00000000+%0A++++%7D+x%3B%0A++++s+%3D+sizeof(x)%3B%0A%7D'),l:'5',n:'1',o:'C+source+%231',t:'0')),k:57.65137043047197,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:cmos-c64-trunk,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'-Os',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+llvm-mos+commodore+64+(Editor+%231)',t:'0')),k:42.34862956952804,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4

mos-platform/geos-cbm/README.md
```diff
@@ -0,0 +1,415 @@
...
+    noinline void SwapMod(uint8_t req_mod_no)
+    {
+	    if (req_mod_no != curModule) {
+		    curModule = req_mod_no;
+		    ReadFile(OVERLAYADDR, OVERLAYSIZE, swapTable[req_mod_no-1]);
+	    }
+    }
+
+`OVERLAYADDR` and `OVERLAYSIZE` are defined by the linker and default to `0x5000` and `0x1000` (for details see the `vlir.ld` linker script in `.../mos-platform/geos-cbm/vlir.ld`).
+You can change the size of the overlay memory to e.g. 2k bytes by adding the following statement to any of your C files:
```
mysterymath, line 353: 0x2000 isn't 2KiB; it's 8KiB. (I do this a lot ;) The thing to remember is that 0x1000 is 4KiB.

mos-platform/geos-cbm/clang.cfg
```diff
@@ -0,0 +1,5 @@
+-mllvm -inline-threshold=0
+-T vlir.ld
+-mlto-zp=107
+-Wl,-linit-stack,--no-check-sections
```
mysterymath, line 4: Something is fishy about `-linit-stack`; merging `common-init-stack` into `libcrt0` `CMakeLists.txt` should ensure that it's always present on the link line without any further intervention. Does the example build without this? If not, what's the error message?

mos-platform/geos-cbm/vlir.ld
```diff
@@ -0,0 +1,218 @@
...
+__overlayaddr 	= 0x8000 - __backbufsize - __overlaysize;
+__stacksize   	= 0x0400;
+__stack   	= __overlayaddr;
+__zp_reg_start  = 0x90;
+__zp_reg_end    = 0xfb;
+__zp_reg_size   = __zp_reg_end - __zp_reg_start;
+
+MEMORY {
+  zp          : ORIGIN = __zp_reg_start, LENGTH = __zp_reg_size
+  vlir00(rwx) : ORIGIN = 0x0400, LENGTH = __overlayaddr - __stacksize - 0x0400
```
mysterymath, line 13: There isn't actually much use to giving more than one of these `(rwx)`; the linker script semantics assign all matching sections to the first match, regardless of whether or not they actually fit. If they don't fit, the link fails. So this mechanism is of very limited utility, and it's usually better to assign sections to memory regions explicity.

mos-platform/geos-cbm/vlir.ld
```diff
@@ -0,0 +1,218 @@
...
+  vlir08(rwx) : ORIGIN = __overlayaddr+7*__overlaysize, LENGTH = __overlaysize
+  vlir09(rwx) : ORIGIN = __overlayaddr+8*__overlaysize, LENGTH = __overlaysize
+  vlir0a(rwx) : ORIGIN = __overlayaddr+9*__overlaysize, LENGTH = __overlaysize
+  vlir0b(rwx) : ORIGIN = __overlayaddr+10*__overlaysize, LENGTH = __overlaysize
+  vlir0c(rwx) : ORIGIN = __overlayaddr+11*__overlaysize, LENGTH = __overlaysize
+  vlir0d(rwx) : ORIGIN = __overlayaddr+12*__overlaysize, LENGTH = __overlaysize
+  vlir0e(rwx) : ORIGIN = __overlayaddr+13*__overlaysize, LENGTH = __overlaysize
+  vlir0f(rwx) : ORIGIN = __overlayaddr+14*__overlaysize, LENGTH = __overlaysize
+
+  dir_block(rw)       : ORIGIN = 0x0000, LENGTH = 0xfe
```
mysterymath, line 30: These shouldn't have type tags, since it doesn't make sense to assign sections to them.

mos-platform/geos-cbm/clang.cfg
mysterymath, line 4: We've used a "trick" to avoid `--no-check-sections`, and I think it should probably work here too.

All llvm-mos LLD linker addresses are 32 bits, regardless of how the address is actually used. Relocations will intentionally silently discard the high two bytes. This allows the high two bytes to encode banking information, and it simplifies linker scripts by giving every section a distinct link-time address.

Conventionally, we've used the highest byte as a "category" or "namespace", and the low byte as a "bank number". In particular, it's fairly straightfoward to access the bank number of an address in assembly; we'd like to extend this to C too once we figure out how to spell it.

Dirk-Schnorpfeil: Hi, i am not really sure that i got your point. As it is a little bit more work, let me ask first:

Are you saying i should start the overlay sections at 0x010000, 0x020000, 0x030000, .... ?

So this change in the code:

SECTIONS { 
  
  INCLUDE c.ld 

  .dir_block : { *(.dir_block .dir_block.*) } >dir_block
  .info_block : { *(.info_block .info_block.*) } >info_block

  .vlir01 __overlayaddr +0x010000 : AT (__overlayaddr+0*__overlaysize) { *(.vlir01 .vlir01.*) NOCROSSREFS(.vlir01 .vlir02 .vlir03 .vlir04 .vlir05 .vlir06 .vlir07 .vlir08 .vlir09 .vlir0a .vlir0b .vlir0c .vlir0d .vlir0e .vlir0f) } > vlir01
  .vlir02 __overlayaddr +0x020000 : AT (__overlayaddr+1*__overlaysize) { *(.vlir02 .vlir02.*) NOCROSSREFS(.vlir01 .vlir02 .vlir03 .vlir04 .vlir05 .vlir06 .vlir07 .vlir08 .vlir09 .vlir0a .vlir0b .vlir0c .vlir0d .vlir0e .vlir0f) } > vlir02
 
...

mysterymath: Yep, that's correct.

mos-platform/geos-cbm/vlir.ld
```diff
@@ -0,0 +1,218 @@
...
+REGION_ALIAS("c_writeable", vlir00)
+
+SECTIONS { 
+  
+  INCLUDE c.ld 
+
+  .dir_block : { *(.dir_block .dir_block.*) } >dir_block
+  .info_block : { *(.info_block .info_block.*) } >info_block
+
+  .vlir01 __overlayaddr : AT (__overlayaddr+0*__overlaysize) { *(.vlir01 .vlir01.*) NOCROSSREFS(.vlir01 .vlir02 .vlir03 .vlir04 .vlir05 .vlir06 .vlir07 .vlir08 .vlir09 .vlir0a .vlir0b .vlir0c .vlir0d .vlir0e .vlir0f) } > vlir01
```
mysterymath, line 45: It should be possible to break out NOCROSSREFS and only specify it once; I don't think it needs to appear in each output section to work.

See https://sourceware.org/binutils/docs/ld/Miscellaneous-Commands.html#index-NOCROSSREFS_0028sections_0029

Dirk-Schnorpfeil: Found no way. Always get error messages if i put it somewhere else.

Can you give me an example where exactly to put the NOCROSSREFS clause?

Thanks a lot!

mos-platform/geos-cbm/geos_constants.h
Dirk-Schnorpfeil: I grepped through common/include. None of these files defines NULL...

Maybe they include some standard clang header which do?

I would really like to have NULL defined in GEOS headers, as its used at several places (see also the *.c in example/geos-cbm)...

mysterymath: NULL is defined in the `stddef.h` chain that comes directly from the clang compiler. Defining it is considered part of the responsibilities of the compiler. In particular, in C++ it's defined as a compiler intrinsic `__null`. (I just learned this today by looking at the source.)

> I would really like to have NULL defined in GEOS headers, as its used at several places (see also the *.c in example/geos-cbm)...

I couldn't see any references in the example files. Is there a technical issue preventing use of `\0` in the contexts where NULL is to be used?

mos-platform/geos-cbm/geos_constants.h
mysterymath: I also quickly ran this example through my local build from tip-of-tree llvm-mos, to the same result.

mos-platform/geos-cbm/vlir.ld
mysterymath: Ah, I see what's happening. `NOCROSSREFS` actually only works at the top level; it needs to be a sibling of `SECTIONS`. You've defined an input section description for the file `NOCROSSREFS` that matches the input sections `.vlir01`, `.vlir02`, etc. This is a place where lld's linker script parser is very weak; it doesn't really prevent keywords from being used out of context.

mos-platform/geos-cbm/geos_constants.h
Dirk-Schnorpfeil: Removed NULL, one can include stddef.h if needed

mos-platform/geos-cbm/geos_crt.c
```diff
@@ -0,0 +1,1881 @@
...
+
+bool TestPoint(uint16_t x, uint8_t y) {
+  bool is_pixel_set;
+
+  __r3 = x;
+  __r11L = y;
+  __attribute__((leaf)) asm volatile(
+      "jsr __TestPoint \n"
+      "ldx #0 \n"
+      "bcc pixel_unset \n"
```
mysterymath, line 42: The use of a named label will cause problems if TestPoint is ever inlined. (This applies generally throughout this file.) Given that the scope is small, it might be better to just use `1f` and `1:`. Otherwise, `%=` can be used to generate a number that is unique for each ASM block in a translation unit; that number can be appended to the label.

mos-platform/geos-cbm/geos_crt.c
```diff
@@ -0,0 +1,1881 @@
...
+
+void GetPtrCurDkNm(disk_name_t buffer) {
+  __attribute__((leaf)) asm volatile("ldx #2 \n"
+                                     "jsr __swap_userzp \n"
+                                     "jsr __GetPtrCurDkNm \n"
+                                     "jsr __swap_userzp " : : : "a",
+                                     "x", "y", "c", "v");
+  uint8_t idx = 0;
+  const char *p = (const char *)__r0;
+  while ((idx < DK_NM_ID_LEN) && (*p != (char)0xa0)) {
```
mysterymath, line 714: nit: no braces for single line while body

mos-platform/geos-cbm/geos_crt.c
```diff
@@ -0,0 +1,1881 @@
...
+                                     "jsr __swap_userzp " : "=x"(errno) : : "a",
+                                     "y", "c", "v");
+
+  return errno;
+}
+
+void MainLoop(void) {
+  __attribute__((leaf)) asm volatile("jsr __MainLoop " : : : "a", "x", "y", "c",
+                                     "v");
+}
```
mysterymath, line 1881: End file with newline.

mos-platform/geos-cbm/crt0.c
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * This file contains the implemantation of the "C" definition of
+ * init and fini code and the directory entry and info sector for
+ * use in GEOS applications  with llvm-mos C compiler.
+ *
+ ***********************************************************************/
+
+#include "geos.h"
+
+asm(".set __zp_reg_size, 0xfb - 0x90 \n"
```
mysterymath, line 12: It's easier to read large blocks of asm like this if in their own `.s` files, since it avoids all the C syntactic cruft.

mos-platform/geos-cbm/crt0.c
```diff
@@ -0,0 +1,102 @@
...
+    "__after_main:\n"
+    "     jsr __swap_userzp\n"
+    "     jmp __EnterDeskTop\n");
+
+void EnterDeskTop(void) {
+  __attribute__((leaf)) asm volatile("jsr __swap_userzp \n"
+                                     "jmp __EnterDeskTop\n");
+}
+
+__attribute__((section(".dir_block")))
```
mysterymath, line 55: nit: `__attribute__((section(...), retain, weak)`
Here and elsewhere; improves syntactic signal-to-noise ratio.

mos-platform/geos-cbm/crt0.c
```diff
@@ -0,0 +1,102 @@
...
+__attribute__((retain))
+__attribute__((weak)) 
+cvt_dir_entry_t __std_dir_entry = {
+    {USR,
```
mysterymath, line 59: Since this opens a struct, the contents should be indented. Did clang-format do that? If so; disregard; it's a losing battle fighting clang-format, unfortunately.

mos-platform/geos-cbm/crt0.c
```diff
@@ -0,0 +1,102 @@
...
+    {0, 0xff},
```
mysterymath, line 60: These struct data items could use comments with the field name for any where it isn't immediately clear from the value provided (e.g. contrast `APPLICATION` and `0`).

examples/geos-cbm/SamVlirHdr.c
```diff
@@ -0,0 +1,43 @@
...
+ ********************************************************************/
+
+#include <geos.h>
+
+#include "SamVlir.h"
+
+__attribute__((section(".info_block"))) 
+__attribute__((retain)) 
+file_header_t __std_file_header = {
+    3,
```
mysterymath, line 16: These struct data items could use comments with the field name for any where it isn't immediately clear from the value provided (e.g. contrast APPLICATION and 3).

mlund:
Please also update the root level README.md to include the geos target.

mos-platform/geos-cbm/crt0.c
Dirk-Schnorpfeil: It was screwed by clang-format, so i keep it this way

examples/geos-cbm/SamVlirHdr.c
Dirk-Schnorpfeil: Its a fixed value (number of bytes for one pixel row of an icon). It is documented at the type (file_header_t) and in the GEOS programming documentation, so i feel, i will not repeat it here...

mos-platform/geos-cbm/crt0.c
Dirk-Schnorpfeil: Documented in the type and GEOS documentation

mos-platform/geos-cbm/crt0.c
Dirk-Schnorpfeil: Used R"ASM(

That should do it for now

README.md
```diff
@@ -165,6 +166,7 @@ executables and libraries for that target.
 | Commodore                       | VIC-20             | `mos-vic20-clang`                |
 | CP/M-65                         | -                  | `mos-cpm65-clang`                |
 | Dodo 6502 Game System           | -                  | `mos-dodo-clang`                 |
+| Commodore GEOS                  | 64, 128            | ˋmos-geos-cbm-clanˋ              |
```
mysterymath, line 169: `cbm-clang`

mysterymath (approved):
Just one last typo, but otherwise, LGTM! Thanks for all your hard work putting this together and bearing with the code review process.

README.md
Dirk-Schnorpfeil: On the road, will fox it tomorrow! Thanks a lot to both of you for your time. I learned a lot!!!

PR #376, opened 2024-09-29
Fix strpbrk()

ivop:
Hi,

I traced back a misbehaving qe (editor in CP/M-65) to strtok() not working correctly, which in turn turned out to be strpbrk(). This fixes that and also results in a slightly smaller binary.

Regards,
Ivo


mysterymath (approved):

mysterymath:
Good catch, thanks!

PR #377, opened 2024-09-30
cd: Build llvm-mos-sdk for Apple Silicon

pr1metine:
Related to llvm-mos/llvm-mos#444. With this PR, the SDK will be bundled alongside an Apple Silicon build of LLVM-MOS. I.e., this PR produces `llvm-mos-macos-arm64.tar.xz`. This should exclusively Mach-O universal binaries, e.g., clang. They should be executable both on x86_64 and AArch64.

Please merge llvm-mos/llvm-mos#464 before this.

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - name: Fetch the latest llvm-mos-sdk prerelease.
         run: |
-          curl -ZLO "https://github.com/llvm-mos/llvm-mos-sdk/releases/download/prerelease/{llvm-mos-linux.tar.xz,llvm-mos-macos.tar.xz,llvm-mos-windows.7z}"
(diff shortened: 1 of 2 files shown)


pr1metine:
Hmm, looks like the existing binary `llvm-mos-macos.tar.xz` is already a Mach-O universal binary.

mysterymath:
> Hmm, looks like the existing binary `llvm-mos-macos.tar.xz` is already a Mach-O universal binary.

Interesting; does it contain both x86-64 and AArch64? If so, has anyone verified whether it runs natively on Apple Silicon?

pr1metine:
Yes and yes. Apparently, Apple Clang produces universal macOS binaries [by default](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary) regardless of the host platform.

Besides, I just checked [the latest MacOS release of llvm-mos-sdk](https://github.com/llvm-mos/llvm-mos-sdk/releases/download/v20.2.0/llvm-mos-linux.tar.xz). [By running `lipo`](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary#Determine-Whether-Your-Binary-Is-Universal), I can confirm that llvm-mos-sdk exclusively contains universal binaries. I was able to run them natively on my Apple Silicon MacBook.

pr1metine:
This makes building a separate binary for Apple Silicon largely obsolete. I will close this PR for now

PR #388, opened 2025-01-07
[cpm65] cpm_bios_seldsk_i() should return NULL on error

eduardocasino:
cpm_bios_seldsk_i() is currently using the default exiter and so returns undefined results on error. That function should return a NULL pointer instead.


mysterymath (approved):
Sorry for letting this one slip through. LGTM!

PR #395, opened 2025-04-28
enable stdio on rp6502 target

rumbledethumps:
File operations just needed __osmaperrno to work. I also added assembly code that optimizes RIA calls for size.


mos-platform/rp6502/clock_getres.c
```diff
@@ -1,18 +1,9 @@
-#include <rp6502.h>
+#include "rp6502.h"
 #include <time.h>
 
+extern int __clock_gettimespec(struct timespec *ts, unsigned char op);
```
mysterymath, line 4: nit: extern is superfluous for function declarations. Here and throughout.

mos-platform/rp6502/clock_gettimespec.c
```diff
@@ -0,0 +1,13 @@
+#include "rp6502.h"
+#include <time.h>
+
+int __clock_gettimespec(struct timespec *ts, unsigned char op)
+/* Internal method shared by clock_getres and clock_gettime. */
+{
```
mysterymath, line 6: Nit: move the comment before the signature and place the brace on the opening line.

mos-platform/rp6502/ria.s
```diff
@@ -0,0 +1,86 @@
+; Helpers for building RIA API shims
```
mysterymath, line 1: Do these need to be in assembly? It seems like having them in C would allow them to be inlined; wouldn't that usually be preferable?

mysterymath (approved):

mos-platform/rp6502/ria.s
rumbledethumps: The C language isn't expressive enough to optimize RP6502 OS calls. By design, this assembly will make smaller and slower code in most cases. Only OS calls use these, which are expensive enough that a few extra cycles aren't a big deal. Anyone wanting faster code won't be using the standard library so they can short stack and ignore errno unless it's truly needed.

mos-platform/rp6502/clock_getres.c
rumbledethumps: Just a habit for future me to know it's really external and not just declared early in the C file.

PR #396, opened 2025-04-28
MMC1 512K PRG support

sebastienvezina:
Added new banks to prg-rom-banked.ld and CHR register bit sets to set_prg_bank


mysterymath:
This actually looks pretty close to correct already; I think we'll be able to whip this into shape without much new work.

I'll take a closer look at this later today.

mos-platform/nes-mmc1/prg-rom-banked.ld
```diff
@@ -15,6 +15,22 @@ __prg_rom_11_lma    = 0x0b8000;
...
+__prg_rom_17_lma    = 0x118000;
+__prg_rom_18_lma    = 0x128000;
+__prg_rom_19_lma    = 0x138000;
+__prg_rom_20_lma    = 0x148000;
+__prg_rom_21_lma    = 0x158000;
+__prg_rom_22_lma    = 0x168000;
+__prg_rom_23_lma    = 0x178000;
+__prg_rom_24_lma    = 0x188000;
+__prg_rom_25_lma    = 0x198000;
+__prg_rom_26_lma    = 0x1A8000;
```
mysterymath, line 29: nit: lowercase hex for consistency

mos-platform/nes-mmc1/prg-rom-banked.ld
```diff
@@ -32,6 +48,22 @@ __prg_rom_11_offset    = 0x2c000;
 __prg_rom_12_offset    = 0x30000;
 __prg_rom_13_offset    = 0x34000;
 __prg_rom_14_offset    = 0x38000;
+__prg_rom_15_offset    = 0x3C000;
```
mysterymath, line 51: same as above; lowercase hex

mos-platform/nes-mmc1/prg-rom-banked.ld
```diff
@@ -15,6 +15,22 @@ __prg_rom_11_lma    = 0x0b8000;
 __prg_rom_12_lma    = 0x0c8000;
 __prg_rom_13_lma    = 0x0d8000;
 __prg_rom_14_lma    = 0x0e8000;
+__prg_rom_15_lma    = 0x0f8000;
```
mysterymath, line 18: As we discovered, there is no PRG-ROM 15. Both 15 and 31 are just 'fixed'.

mos-platform/nes-mmc1/prg-rom-banked.ld
```diff
@@ -51,6 +83,22 @@ MEMORY {
   prg_rom_12    : ORIGIN = __prg_rom_12_lma,    LENGTH = __prg_rom_size >= 256 ? 0x4000 - 12 : 0
   prg_rom_13    : ORIGIN = __prg_rom_13_lma,    LENGTH = __prg_rom_size >= 256 ? 0x4000 - 12 : 0
   prg_rom_14    : ORIGIN = __prg_rom_14_lma,    LENGTH = __prg_rom_size >= 256 ? 0x4000 - 12 : 0
+  prg_rom_15    : ORIGIN = __prg_rom_15_lma,    LENGTH = __prg_rom_size >= 512 ? 0x4000 - 12 : 0
```
mysterymath, line 86: There is no prg_rom_15; this is prg_rom_fixed. Remove this line.

mos-platform/nes-mmc1/prg-rom-banked.ld
```diff
@@ -78,6 +126,22 @@ SECTIONS {
   .prg_rom_12    : { *(.prg_rom_12    .prg_rom_12.*)    } >prg_rom_12
   .prg_rom_13    : { *(.prg_rom_13    .prg_rom_13.*)    } >prg_rom_13
   .prg_rom_14    : { *(.prg_rom_14    .prg_rom_14.*)    } >prg_rom_14
+  .prg_rom_15    : { *(.prg_rom_15    .prg_rom_15.*)    } >prg_rom_15
```
mysterymath, line 129: Ditto.

mos-platform/nes-mmc1/prg-rom-banked.ld
```diff
@@ -88,7 +152,7 @@ SECTIONS {
   .reset_fixed : { KEEP(*(.reset_fixed)) } >reset_fixed
 }
 
-PROVIDE(__prg_rom_size = 256);
+PROVIDE(__prg_rom_size = 512);
```
mysterymath, line 155: We should still default to 256 I think. We've tended to set our defaults to the most likely configuration, not the most capable. (Although among the most likely configurations, we prefer the most capable.)

mos-platform/nes-mmc1/prg-rom-banked.ld
```diff
@@ -108,5 +172,23 @@ OUTPUT_FORMAT {
...
+  FULL(prg_rom_22)    FULL(reset, 0, LENGTH(prg_rom_22) ? LENGTH(reset) : 0)
+  FULL(prg_rom_23)    FULL(reset, 0, LENGTH(prg_rom_23) ? LENGTH(reset) : 0)
+  FULL(prg_rom_24)    FULL(reset, 0, LENGTH(prg_rom_24) ? LENGTH(reset) : 0)
+  FULL(prg_rom_25)    FULL(reset, 0, LENGTH(prg_rom_25) ? LENGTH(reset) : 0)
+  FULL(prg_rom_26)    FULL(reset, 0, LENGTH(prg_rom_26) ? LENGTH(reset) : 0)
+  FULL(prg_rom_27)    FULL(reset, 0, LENGTH(prg_rom_27) ? LENGTH(reset) : 0)
+  FULL(prg_rom_28)    FULL(reset, 0, LENGTH(prg_rom_28) ? LENGTH(reset) : 0)
+  FULL(prg_rom_29)    FULL(reset, 0, LENGTH(prg_rom_29) ? LENGTH(reset) : 0)
+  FULL(prg_rom_30)    FULL(reset, 0, LENGTH(prg_rom_30) ? LENGTH(reset) : 0)
+  FULL(prg_rom_fixed) FULL(reset_fixed)
```
mysterymath, line 191: Like the earlier sections, we should only include this if the size is actually over 512. We can't just check whether `prg_rom_fixed` has a non-zero size though, since it always will. Accordingly, we'll need to use expression like `__prg_rom_size >= 512 ? 0x4000 - 12 : 0` here for both `prg_rom_fixed` and `reset_fixed`.

The syntax is `FULL(<section>, <start>, <length>)` to trim a section's output as it's added to the ROM file. The `<start>` should always be 0, but the `<length>` should either be 0 or the size of the section (`LENGTH(<section>)`), depending on whether or not we have 512 PRG-ROM.

mos-platform/CMakeLists.txt
```diff
@@ -85,4 +85,4 @@ add_subdirectory(pce)
 add_subdirectory(pce-cd)
 add_subdirectory(rp6502)
 add_subdirectory(supervision)
-add_subdirectory(vic20)
+# add_subdirectory(vic20)
```
mysterymath, line 88: nit: remove

.gitignore
```diff
@@ -5,3 +5,4 @@ build/**
 build*
 *~
 *.rej
+.qodo
```
mysterymath, line 8: nit: remove

mos-platform/nes-mmc1/prg-rom-banked.ld
```diff
@@ -90,6 +150,8 @@ SECTIONS {
 
 PROVIDE(__prg_rom_size = 256);
 
+prg_rom_is_512 = __prg_rom_size >= 512 ? 1 : 0;
```
mysterymath, line 153: Nit: this symbol needs to be prefixed `__` to meet the contract between a C hosted environment and a standards compliant C program. (Such symbol names are reserved by the C standard for use by the toolchain, i.e., us.)

mos-platform/nes-mmc1/mapper.s
```diff
@@ -99,23 +99,45 @@ get_prg_bank:
...
+  rts
 .Lretry:
-	tya
-	jmp .Lset
-
-
+  tya
+  jmp .Lset
+.Luse_suxrom_bankswitching:
+  .byte ((<prg_rom_is_512) << 7)
```
mysterymath, line 140: This is extremely fishy; there shouldn't be any way that the linker can actually fill in this value. It can't actually read expressions like `(<prg_rom_is_512) << 7` out of the object file for it to resolve them; such expressions can only be resolved by the assembler. But there's no way the assembler can know the value of `prg_rom_is_512`. Does this successfully compile? If so, can you paste the ELF file on Discord? I need to take a look at it; there may be an assembler bug here.

mos-platform/nes-mmc1/prg-rom-banked.ld
```diff
@@ -107,6 +169,24 @@ OUTPUT_FORMAT {
   FULL(prg_rom_12)    FULL(reset, 0, LENGTH(prg_rom_12) ? LENGTH(reset) : 0)
   FULL(prg_rom_13)    FULL(reset, 0, LENGTH(prg_rom_13) ? LENGTH(reset) : 0)
   FULL(prg_rom_14)    FULL(reset, 0, LENGTH(prg_rom_14) ? LENGTH(reset) : 0)
-  FULL(prg_rom_fixed) FULL(reset_fixed)
+  FULL(prg_rom_fixed) FULL(reset_fixed, 0, __prg_rom_size >= 512 ? 0x4000 - 12 : 0)
```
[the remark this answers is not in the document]
sebastienvezina: Updated

mos-platform/nes-mmc1/mapper.s
```diff
@@ -99,9 +99,39 @@ get_prg_bank:
...
+	lda _CHR_BANK0
+	eor #%00010000
+	jsr set_chr_bank_0_retry
+	lda _CHR_BANK1
+	eor #%00010000
+	jsr set_chr_bank_1_retry
+.Lset_inner_bank:
+  	; restore the bank byte and put it in y so we can reload it on retry
+  	pla
+.Lcontinue_bank_switch:
```
mysterymath, line 122: This leaves no way to enable/disable the PRG-RAM. The previous version of this code was very hands-off about the nature of the values written to CHR-BANK and PRG-BANK. That's kind of intentional: I didn't have a good plan for how to efficiently set a semantic bank number across the wild set of CHR and PRG bank regs in MMC1, especially across the different variants. We could definitely do so, but it seems... maybe safer to advertise that these are uninterpreted values, and that if you want to set a PRG bank to the high ones, you need to manipulate the CHR bank as you would do by hand.

I think this is dramatically less ergonomic than something nicer, but this change represents a mixed intermediate state that seems worse than being either fully-hands on or fully hands off. And it seems valuable to have this in the meantime, so I'd request restoring this to its older shape (presuming that is still usable.)

mysterymath (commented):
Woooow, sorry for dropping this one on the floor. It's been a year lol. Yikes. I think this can go in, sans one bit of feedback.

PR #406, opened 2025-08-12
nes-unrom: Fix banked_call()

### LINKED ISSUE #405: Potential bug in nes-unrom's banked_call()
The implementation of banked_call saves the original bank in the `y` register, but that being a caller-saved register, the function being called is free to modify it, destroying the original bank information.

wendelscardua:
Fixes #405 

Saves original bank to the hardware stack instead of using the y register


mos-platform/nes-unrom/mapper.s
```diff
@@ -10,12 +10,16 @@
 .section .text.banked_call,"ax",@progbits
 .weak banked_call
 banked_call:
-	ldy _BANK_SHADOW
+    tay
+	lda _BANK_SHADOW
+	pha
+	tya
 	jsr set_prg_bank
+	pha
```
mysterymath, line 18: This adds two `pha` but only a single `pla`, which definitely seems wrong.

mysterymath (approved):

PR #407, opened 2025-08-28
Neo6502 platform API: controller status API

patrick-flynn:
file: `mos-platform/neo6502/api/controller.c` [link](https://github.com/llvm-mos/llvm-mos-sdk/blob/main/mos-platform/neo6502/api/controller.c)
function:` uint8_t neo_controller_read(uint8_t index)`
bug 1: return type should be `uint32_t`
bug 2: `index` is not placed in `ControlPort.params[0]` before the API is invoked.


asiekierka:
LGTM

PR #410, opened 2025-09-07
Stop __set_heap_size from trashing soft stack

johnwbyrd:
Limit the __set_heap_size function to not trash the stack if too much memory is requested.  Additionally, choose a default stack size and use it as a safety margin when increasing heap size.

The new stack size, __stack_reserve_size, is set weakly to 1024, and anyone who wants to change it can do so without recompiling.

Additionally, the simulator now flushes on stdout output.  This helps establish that long-running tasks are still running on mos-sim.

These changes were necessary to be able to get pi.c to calculate to 50000 digits on the Commodore 64.

### DIFF
--- a/mos-platform/common/c/malloc.cc
+++ b/mos-platform/common/c/malloc.cc
@@ -16,6 +16,20 @@
 
 extern char __heap_start;
 extern char __heap_default_limit;
+extern char __stack_reserve_size;
+
+// Get current soft stack pointer value
+__asm__ (
+".text\n"
+".global __get_current_stack_top\n"
+"__get_current_stack_top:\n"
+"  lda __rc0\n"    // Load low byte of soft stack pointer
+"  ldx __rc1\n"    // Load high byte of soft stack pointer  
+"  rts\n"
+);
+
+// Declare the assembly function for C
+extern "C" size_t __get_current_stack_top();
 
 namespace {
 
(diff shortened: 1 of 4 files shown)


johnwbyrd:
With this change, a user can:

```
__set_heap_limit(__get_heap_max_safe_size());
```

and they will get the maximum amount of platform memory available for malloc().

mos-platform/common/c/malloc.cc
```diff
@@ -179,6 +193,7 @@ FreeChunk *find_fit(size_t size) {
   return nullptr;
 }
 
+"  rts\n"
```
mysterymath, line 196: This is definitely a typo.

mos-platform/common/c/malloc.cc
```diff
@@ -16,6 +16,20 @@
 
 extern char __heap_start;
 extern char __heap_default_limit;
+extern char __stack_reserve_size;
+
+// Get current soft stack pointer value
+__asm__ (
+".text\n"
+".global __get_current_stack_top\n"
```
mysterymath, line 24: We can avoid making this a global symbol at all by using inline assembly in a static function. This would also allow the compiler to inline this, which would usually be beneficial.

mysterymath (approved):
Thanks, this is a really nice change! Few comments, otherwise LGTM!

mysterymath (approved):
LGTM (made the edit myself lol)

PR #415, opened 2025-10-29
[SNES] Add target

Phillip-May:
Here's an initial draft for a snes target, based my snes xc libraries, so registers are named with REG_ for example REG_INIDISP, the example provided is an input test that changes the screens colour. I've already used essentially this target for my demake/port of c celeste.


mos-platform/snes/vectors.s
```diff
@@ -0,0 +1,301 @@
+; SNES Interrupt Vector Handlers for LLVM-MOS (6502 compatible)
+; Based on vectors.asm reference from wdc816cc
```
asiekierka: We can't really "base" code on WDC's, unfortunately. I assume the licenses are not compatible.

However, what little potentially copyrightable material I see in this file requires a rework anyway...

mos-platform/snes/clang.cfg
```diff
@@ -0,0 +1,2 @@
+-mlto-zp=224
```
asiekierka: `-mcpu=mosw65816` will enable some 65816-specific optimizations.

PR #418, opened 2025-12-03
added __appram_start parameter

Dirk-Schnorpfeil:
Some GEOS application do not start at the default address 0x0400. Instead the developer can now define the start address and tell through the SET_APPRAM_START macro.

Anyhow, developer must not forget to give the start address also in the programs info sector.


mysterymath:
LGTM, thanks!

PR #420, opened 2026-01-06
[C128] Add missing initialization of FNAM_BANK ($C7) to 0.

swildner:
This allows open() to work out of the box instead of failing with an error 62 (file not found).

cc65's libsrc/c128/crt0.s does it like this as well.


swildner:
I'm not sure though if init-mmu.S is the best place to add this.

mysterymath:
> I'm not sure though if init-mmu.S is the best place to add this.

Sorry for not getting around to this. I think ideally this would go in its own file. That file can export a global symbol, something like `__init_FNAM_BANK`, that other files could reference (just by including the line `.globl __init_FNAM_BANK`) if they depend on that value being initialized for their operation. Then, that initialization would be performed if and only iff some aspect of the SDK that depended on them was used. Otherwise, the user can freely discharge this requirement and do their own shiz.

PR #439, opened 2026-06-02
Add INTERMEDIATE platform kind; hide atari8-common wrappers

cwedgwood:
## What

Adds an `INTERMEDIATE` option to the `platform()` function in `cmake/platform.cmake`. When set, the platform's library/cfg are still built and installed, but the `mos-<name>-clang` / `clang++` / `clang-cpp` driver wrappers are not.

Applies it to `atari8-common`.

## Why

`atari8-common` exists only as a shared layer that `atari8-dos`, `atari8-cart-std`, `atari8-cart-megacart` and the `atari5200-*` carts chain through via `@mos-atari8-common.cfg`. There is no "atari8-common" machine, so the installed `mos-atari8-common-clang*` wrappers are misleading: a user invoking one gets a successful link that cannot run on real hardware.

## Scope

Install-side only. No library content changes. The generated `mos-atari8-common.cfg` is still installed so descendant `@`-includes resolve unchanged.

## Verification

- Built and installed against a fresh `/store0/opt/llvm-mos`.
- Post-install: `mos-atari8-common-clang*` absent; `mos-atari8-common.cfg` present.
- All atari8 descendant wrappers (`mos-atari8-dos-clang`, `mos-atari8-cart-std-clang`, ...) present and produce valid `.xex` / `.car` output.
- `test/atari8-dos` (2/2) and `test/atari8-cart-std` (7/7) pass.

## Follow-ups (not in this PR)

Other `HOSTED PARENT` platforms that look like analogous "shared layer" candidates and could plausibly become `INTERMEDIATE` in future PRs: `pce-common`, `nes-common`, `commodore`. Out of scope here.

### DIFF
--- a/cmake/platform.cmake
+++ b/cmake/platform.cmake
@@ -29,7 +29,7 @@ endfunction()
 # present. A HOSTED platform supports the fragment of the C standard library
 # currently implemented. A COMPLETE platform cannot have any descendants.
 function(platform name)
-  cmake_parse_arguments(ARGS "COMPLETE;HOSTED" PARENT "" ${ARGN})
+  cmake_parse_arguments(ARGS "COMPLETE;HOSTED;INTERMEDIATE" PARENT "" ${ARGN})
 
   set(PLATFORM ${name})
   set(PLATFORM ${name} PARENT_SCOPE)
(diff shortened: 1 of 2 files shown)


mysterymath (changes requested):
These are present as a feature to enhance portability, see: https://llvm-mos.org/wiki/Porting

These can be used along with other files in the user project to build against targets not yet directly supported by the SDK. At the most abstract, users can build against the common target to support arbitrary 6502 targets, but with some library support. Notably, it's important that this be possible *without* diving into CMake, particularly *llvm-mos-sdk* CMake, which is kinda intrinsically going to be difficult.

That being said, I am receptive to the idea that this may be confusing to users, and that they likely outnumber those that are trying to work on unusual hardware. Still, we shouldn't make the that task as difficult as this patch would; I'm open to other ideas on this.

cwedgwood:
Want me to just make this a docs update then? Or drop it entirely?

mysterymath:
If you can find a way to make it clearer in the docs yeah, but I'd also be open to having a way to *gate* these targets somehow... No idea what that would look like though. We could also maybe have a better naming convention for these... Maybe base rather than common? Something that implies incompleteness...

But yeah, this is not gonna fly as is.

PR #441, opened 2026-06-22
RP6502 0.28 support, add argv, time support

rumbledethumps:


mos-platform/rp6502/mainargs.c
```diff
@@ -0,0 +1,56 @@
...
+#include <stdint.h>
+
+int __argc;
+char **__argv;
+
+/* Override to provide storage for argv, e.g.:
+ *   void *argv_mem(size_t size) { return malloc(size); }
+ * Returns NULL (default) to silently skip argc/argv setup.
+ */
+__attribute__((weak)) void *argv_mem(size_t size) {
```
mysterymath: This needs to be `__argv_mem`, since this file is unconditionally included in user code, and it accordingly may conflict with a user-defined symbol. A user's intent to override needs to be by placing something in the C reserved namespace in such cases.

mysterymath: No action required, but it might be nice to provide a `-l` library that the user could insert on the command line that would provide a version of this that works with `malloc`. That would allow users to simply opt-in to the expense of argv handling.

I think there's also another way to deal with this: I think one could do some SP trickery to place argv on the soft stack. This would run the risk of overflow, but it would save any need for malloc. Up to you.

mysterymath (approved):

mos-platform/rp6502/mainargs.c
rumbledethumps: There's not always room for 512 bytes of possible argv. A function can refuse to provide the memory. I also want to maintain parity with cc65. This solution isn't elegant, but it does tick all the boxes. It's unlikely this will bleed into other platforms since the RP6502 is doing things here no other platform does.

PR #444, opened 2026-07-29
[apple2] Add new Apple II ProDOS target

calcwatch:
## Summary

Add a hosted `apple2` platform targeting ProDOS 8 system programs.

The target:

- Produces ProDOS `SYS` binaries loaded at `$2000`.
- Supports 64 KiB Apple II systems and Apple IIe systems in 40-column mode.
- Provides console input and output through the Monitor ROM firmware.
- Initializes the standard keyboard and display firmware vectors during startup.
- Returns to the ProDOS dispatcher with the ProDOS `QUIT` MLI call.
- Reserves ProDOS memory and zero-page regions in the linker configuration.
- Ensures executable code is placed first in the SYS load image, ahead of initialized zero-page data.

Hi-res graphics and auxiliary-memory support are intentionally out of scope for this initial target.

## Testing

- `cmake --build build --target test-apple2`
  - Minimal hosted program compilation passes.
  - Oversized RAM image rejection passes.
- Manually tested under ProDOS 8 in MAME.
  - Verified formatted text output.
  - Verified blocking keyboard input and character echo.
  - Verified returning to the ProDOS system selector after program exit.


mos-platform/apple2/link.ld
```diff
@@ -0,0 +1,44 @@
...
+ */
+SECTIONS {
+    INCLUDE text.ld
+    INCLUDE rodata.ld
+    INCLUDE data.ld
+    INCLUDE zp.ld
+    INCLUDE bss.ld
+    INCLUDE noinit.ld
+}
+ASSERT(_start == ORIGIN(ram), "ProDOS SYS entry point must be $2000.")
```
mysterymath: This is more fragile than having a dedicated `.sys_entry` section containing `JMP _start` placed before everything in `ram`. This allows general C sections can be placed wherever, and it also appears to be a general ProDOS convention (forming part of the interpreter protocol, which sets us up for that later).

mos-platform/apple2/_Exit.S
```diff
@@ -0,0 +1,46 @@
+.global _Exit
+
+; Establish the state required of a ProDOS system program. Keep this in the
+; same object as _Exit so archive extraction always includes it.
+.section .init.050,"ax",@progbits
```
mysterymath: Typically such code belongs in `crt0.o`, which is always included. The current `crt0.o` is falling through from the parent `common` target, but the SDK provides means to declare an apple2-specific one with both those contents and e.g. this routine.

mos-platform/apple2/getchar.c
```diff
@@ -0,0 +1,13 @@
...
+
+__attribute__((always_inline, weak)) int __to_ascii(void *ctx,
+                                                    int (*read)(void *ctx)) {
+  int c = read(ctx);
+  return c == '\r' ? '\n' : c;
+}
+
+int __getchar(void) {
+  unsigned char c;
+  asm volatile("jsr $fd0c" : "=a"(c) : : "x", "y", "p");
```
mysterymath: This follows the llvm-mos calling convention precisely, so it's conventional to call it as if it were a C function.

mos-platform/apple2/apple2.h
```diff
@@ -0,0 +1,10 @@
+#ifndef _APPLE2_H
+#define _APPLE2_H
+
+#include <stdint.h>
+
+#define APPLE2_PRODOS_MLI ((uintptr_t)0xbf00)
+#define APPLE2_PRODOS_VERSION (*(volatile uint8_t *)0xbffd)
```
mysterymath: This is defined as `PRODOS_VERSION` here, and `SYSTEM_VERSION` in the `.inc` file. They should be consistent, in the absence of a strong historical convention or backwards-compatibility concern.

mos-platform/apple2/_Exit.S
```diff
@@ -0,0 +1,46 @@
...
+  ; input/output vectors point into the program image that ProDOS just
+  ; replaced. Restore the Monitor ROM's keyboard and 40-column display links
+  ; (the firmware equivalents of IN#0 and PR#0) before using RDKEY or COUT.
+  jsr $fe89
+  jsr $fe93
+
+.section .text._Exit,"ax",@progbits
+_Exit:
+  ; Invalidate the power-up byte before returning to ProDOS's dispatcher.
+  inc $03f4
```
mysterymath: Do we need do to anything with the power-up byte on entry?

mos-platform/apple2/link.ld
calcwatch: Agreed. I added a dedicated `.sys_entry` section at `$2000` containing `JMP _start` and placed it before the normal C sections. I also replaced the `_start == $2000` assertion with assertions on the entry section itself.

mos-platform/apple2/_Exit.S
calcwatch: I see. I've moved it to `crt0.o`.

mos-platform/apple2/getchar.c
calcwatch: Oh nice, I didn't realize that. Fixed

mos-platform/apple2/apple2.h
calcwatch: Good catch. I renamed this to `APPLE2_PRODOS_SYSTEM_VERSION` so that it matches the `PRODOS_SYSTEM_VERSION` in the `.inc` file.

mos-platform/apple2/_Exit.S
calcwatch: Yes, the ProDOS Technical Reference says that a system program should set the reset vector and fix the power-up byte after gaining control, then invalidate the power-up byte before `QUIT`. I now install an Apple II reset handler at `$03F2/$03F3` and set `$03F4` to the vector high byte XOR `$A5`. The handler reinitializes the hardware stack and enters the ProDOS quit path; `_Exit` still invalidates `$03F4` immediately before `QUIT`.

mos-platform/apple2/link.ld
```diff
@@ -24,19 +24,23 @@ REGION_ALIAS("c_writeable", ram)
...
- * $2000. Keep .text first in the load image. In particular, the generic c.ld
- * order puts the load image for initialized zero-page data before .text,
- * which would make ProDOS execute that data as code.
+ * $2000. Keep the three-byte entry stub first in the load image, followed by
+ * the normal C sections. In particular, initialized zero-page data must not
+ * precede the entry stub in the output file.
  */
 SECTIONS {
+    .sys_entry : { KEEP(*(.sys_entry)) } >ram
     INCLUDE text.ld
```
mysterymath: All of this can reduce down to `INCLUDE c.ld`.

mysterymath (approved):
LGTM pending one remaining nit!

mos-platform/apple2/link.ld
calcwatch: Fixed

PR #445, opened 2026-08-18
[MEGA65] Correct Mega65 headers for VIC and ethernet controller

mlund:
Two mega65 sdk header fixes found while writing a driver. Checked against the VHDL core for MEGA65.

#### `D02F` knocks

`$D02F` selects which registers appear at `$D000-$DFFF` and takes three two-byte sequences according to the MEGA65 core (`viciv.vhdl:2492-2510`): `$A5,$96` for VIC-III, `$47,$53` for VIC-IV, `$45,$54` for the 45E100 buffers. They had no names, so callers wrote bare numbers. Adds `VIC4_KEY_*`.

`_45E100.h` said the controller is "Enabled by writing 0x53 and then 0x47" — the VIC-IV sequence with its bytes reversed, and not what maps the ethernet buffers.

#### Ethernet 45E100 register header

- Masks did not say which register they belong to. 20 in flat enum, and the values repeat: `ETH_RST_MASK` and `ETH_NOPROM_MASK` are both `0b00000001`, in `$D6E0` and `$D6E5`. Now grouped per register.
- The size assertion never ran for C, being inside `#ifdef __cplusplus` — and C callers are the ones most likely to index past the struct. Also establishes that the unaligned `uint16_t miimv` at offset 7 gets no padding.
- `$D6EF` was outside the struct, so the debug window had no name.
- `ETH_RXBF_MASK` described only read. Writing bit 1 of `$D6E1` hands back the receive buffer and presents the next frame, edge-triggered (`ethernet.vhdl:1787`). Adds `ETH_RXROTATE_MASK`.
- `ETH_TXPH`/`ETH_RXPH` are two-bit fields, so ORing the mask sets both bits rather than a value. Adds the shifts.

Also logs that `iomap.txt` from the MEGA65 core annotates `$D6E5.0` as both "disable promiscuous mode" and "enable filtering of unicast frames if MAC address does not match", and `$D6E5.5` as both the multicast and the unicast enable.

Claude Opus 5 was used to check header correctness against the MEGA65 VHDL core files. @mlund reviewed and tested.

### DIFF
--- a/mos-platform/mega65/_vic4.h
+++ b/mos-platform/mega65/_vic4.h
@@ -315,6 +315,22 @@ struct __vic4 {
 static_assert(sizeof(__vic4) == 0x80);
 #endif
 
+/// Values written to VICIV.key to choose which registers appear at
+/// 0xD000-0xDFFF. Write the two bytes of a pair in that order; any other
+/// value returns to VIC-II.
+enum
+#ifdef __clang__
+    : uint8_t
+#endif
+{
+  VIC4_KEY_VICIII_A = 0xa5, //!< ...then VIC4_KEY_VICIII_B for C65/VIC-III
+  VIC4_KEY_VICIII_B = 0x96,
+  VIC4_KEY_VICIV_A = 0x47, //!< ...then VIC4_KEY_VICIV_B for MEGA65/VIC-IV
+  VIC4_KEY_VICIV_B = 0x53,
+  VIC4_KEY_ETH_A = 0x45, //!< ...then VIC4_KEY_ETH_B for the 45E100 buffers
+  VIC4_KEY_ETH_B = 0x54,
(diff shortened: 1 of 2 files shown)


mos-platform/mega65/_vic4.h
```diff
@@ -315,6 +315,22 @@ struct __vic4 {
 static_assert(sizeof(__vic4) == 0x80);
 #endif
 
+/// Values written to VICIV.key to choose which registers appear at
+/// 0xD000-0xDFFF. Write the two bytes of a pair in that order; any other
+/// value returns to VIC-II.
+enum
+#ifdef __clang__
```
mysterymath, line 322: Uh, it's the llvm-mos SDK. It's clang. :P

PR #447, opened 2026-08-20
[MEGA65] Make headers host includable in C/C++

mlund:
Make mega65 headers compile in C/C++ on both target and host. Useful for running host side tests.
Sits on top of #445 and should be merged after that.


mysterymath:
I'm... not sure that this is something we actually want. We haven't done this for any of the other pieces of the SDK, and it doesn't really seem... all that desirable? Unless there's something MEGA-65 specific about this? Can you provide more details about the use case here?

